PR #364, opened 2020-03-06
chore: capture test logs and save as CircleCI artifacts

Shesekino:
store logs as CircleCI artifact after KinD integration test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,9 @@ jobs:
     steps:
       - checkout
       - run:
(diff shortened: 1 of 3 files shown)


Shesekino:
test

Shesekino:
I view it as an environmental concern - CircleCI creates the dir, CircleCI stores the artifacts.
you can have the logs when running locally, but you need to opt-in (so we don't spam files in peoples' work stations).
the console.log in dumpLogs() hints towards how to do that :)

PR #370, opened 2020-03-09
chore/e2e operator

ivanstanev:

### What this does

Adds a new deployment flow - using an Operator - and uses it with OpenShift 4 integration tests.
In order to match the flow of installing an Operator through the UI, we do the following manual steps:
- generate OLM (Operator Lifecycle Manager) files that specify how to deploy the Operator
- apply the SnykMonitor CustomResource - this makes the Operator install a snyk-monitor in the namespace

This same flow would be performed by any user wishing to install the snyk-monitor outside of using the UI so we can be confident it is accurate and covering both manual and UI install.

### More information

- [Jira ticket RUN-684](https://snyksec.atlassian.net/browse/RUN-684)


test/helpers/kubectl.ts
```diff
@@ -67,6 +67,12 @@ export async function createDeploymentFromImage(name: string, image: string, nam
   console.log(`Done Letting Kubernetes decide how to manage image ${image} with name ${name}`);
 }
 
+export async function deleteResource(kind: string, name: string, namespace = 'default') {
```
Shesekino: I'd rather avoid default values (no pun intended)

test/setup/index.ts
```diff
@@ -9,6 +9,7 @@ import * as waiters from './waiters';
 
 const testPlatform = process.env['TEST_PLATFORM'] || 'kind';
 const createCluster = process.env['CREATE_CLUSTER'] === 'true';
+const deploymentType = process.env['DEPLOYMENT_TYPE'] || 'YAML';
```
Shesekino: I don't want to block on this - but I'd love it if you could push for paying back some technical debt around these areas in the next sprint. about one day's work at most?
1. unify the setup parameters under one object:
```
interface setupConfiguration = {
  platform: string;
  createCluster: bool;
  deploymentType: string;
}
```
or something like that.

2. find a more elegant way to inject these values than having 3 env variables blocking our package.json

what do you think?

.circleci/config.yml
```diff
@@ -164,9 +169,17 @@ jobs:
           command: |
             export IMAGE_TAG=$([[ "$CIRCLE_BRANCH" == "staging" ]] && echo "staging-candidate" || echo "discardable")
             export SNYK_MONITOR_IMAGE_TAG="${IMAGE_TAG}-${CIRCLE_SHA1}"
-            export SNYK_OPERATOR_VERSION="0.0.1"
+            export SNYK_OPERATOR_VERSION="0.0.1-${CIRCLE_SHA1}"
             export SNYK_OPERATOR_IMAGE_TAG="${SNYK_MONITOR_IMAGE_TAG}"
             ./scripts/package-operator.sh "${SNYK_OPERATOR_VERSION}" "${SNYK_OPERATOR_IMAGE_TAG}" "${SNYK_MONITOR_IMAGE_TAG}"
+      - run:
+          name: REMOVE TEMPLATED OPERATOR BEFORE PERSISTING TO WORKSPACE
```
Shesekino: let's start using lower cases in these steps?

.circleci/config.yml
```diff
@@ -164,9 +169,17 @@ jobs:
...
+          command: |
+            rm -r snyk-operator/deploy/olm-catalog/snyk-operator/0.0.0
```
Shesekino: I usually prefer ```-f``` here just to avoid flakiness or errors. would that make sense here?

.circleci/config.yml
```diff
@@ -177,6 +190,24 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: snyk-operator
+      - run:
+          name: Install operator-courier
+          command: pip3 install operator-courier
+      - run:
+          name: Upload Operator to Quay
+          command: |
+            export QUAY_TOKEN=$(curl -H "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d "{\"user\": {\"username\": \"${QUAY_USERNAME}\", \"password\": \"${QUAY_PASSWORD}\"}}" | jq -r .token)
```
Shesekino: consider using a script (Python or Shell) to do this, I'm concerned we're loading too many things on our circleci config file

.circleci/config.yml
```diff
@@ -164,9 +169,17 @@ jobs:
           command: |
             export IMAGE_TAG=$([[ "$CIRCLE_BRANCH" == "staging" ]] && echo "staging-candidate" || echo "discardable")
             export SNYK_MONITOR_IMAGE_TAG="${IMAGE_TAG}-${CIRCLE_SHA1}"
-            export SNYK_OPERATOR_VERSION="0.0.1"
+            export SNYK_OPERATOR_VERSION="0.0.1-${CIRCLE_SHA1}"
```
Shesekino: this is what's using the ```upload_operator``` step?

test/helpers/kubectl.ts
```diff
@@ -67,6 +67,12 @@ export async function createDeploymentFromImage(name: string, image: string, nam
   console.log(`Done Letting Kubernetes decide how to manage image ${image} with name ${name}`);
 }
 
+export async function patchResourceFinalizers(kind: string, name: string, namespace = 'default') {
```
Shesekino: please no default values where can be avoided

test/helpers/kubectl.ts
```diff
@@ -102,6 +108,31 @@ export async function getPodLogs(podName: string, namespace: string): Promise<an
   return logsOutput.stdout;
 }
 
+export async function waitForPodsReady(partialName: string, namespace: string): Promise<void> {
```
Shesekino: this seems a bit funky, can we be a bit more clear here? why do we need a partial match followed by waiting for it to be ready? can't we wait for a condition on some deployment to be 1/1 or something like that?

test/setup/deployers/operator.ts
```diff
@@ -0,0 +1,33 @@
...
+  _imageOpts: {
+    imageNameAndTag: string;
+    imagePullPolicy: string;
+  },
+): Promise<void> {
+  const namespace = 'snyk-monitor';
+  await kubectl.createNamespace(namespace);
+
+  const secretName = 'snyk-monitor';
+  const gcrDockercfg = process.env['GCR_IO_DOCKERCFG'] || '{}';
```
Shesekino: this line seems familiar, maybe there's room for this to be consolidated somewhere?

test/setup/index.ts
ivanstanev: Sure, let's do it with some YAML config! 🙂 

Agreed, will be happy to allocate some time to resolve this

.circleci/config.yml
ivanstanev: Not sure what is the question, the upload_operator step matches this versioning scheme!
The operator-sdk and Quay don't allow us to use just a tag, but something that resembles x.y.z-tag is OK

test/setup/deployers/operator.ts
ivanstanev: I think we can move it one level up, in setup.ts, because it's a prerequisite to deploying the monitor itself! It will be identical for every platform, regardless of the deployment type. Would you be OK if I did it in a separate PR? It's not a small change, because if I move this code then a lot of other code needs to be trimmed down (unused vars, the interface for the deploy() function gets affected...)?

.circleci/config.yml
ivanstanev: I think it makes sense to extract it out to some script so it can be run locally, but what is your bigger vision here? To have every CI step be in its own separate script?

.circleci/config.yml
ivanstanev: 🤷‍♂ So far no issues without `-f` but I'm happy to add it anyway!

test/fixtures/operator/installation.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: OperatorGroup
+metadata:
+  name: snyk-operator
+  namespace: snyk-monitor
+spec:
+  targetNamespaces:
+  - snyk-monitor
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
```
shaimendel: I really have no idea what this means, but it looks professional

test/fixtures/operator/installation.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: snyk-operator
+  namespace: snyk-monitor
+spec:
+  targetNamespaces:
+  - snyk-monitor
+---
+apiVersion: operators.coreos.com/v1alpha1
```
shaimendel: is this supposed to be `v1alpha1`?

test/helpers/kubectl.ts
```diff
@@ -67,6 +67,12 @@ export async function createDeploymentFromImage(name: string, image: string, nam
   console.log(`Done Letting Kubernetes decide how to manage image ${image} with name ${name}`);
 }
 
+export async function deleteResource(kind: string, name: string, namespace: string) {
```
shaimendel: shouldn't all other delete* functions use this one?

test/setup/index.ts
shaimendel: not sure if this still needs to be done or not, either as a ticket or a commit :)

.circleci/config.yml
shaimendel: I think we should add a general task to clean our circleci config, it is getting huge

test/helpers/kubectl.ts
```diff
@@ -102,6 +108,22 @@ export async function getPodLogs(podName: string, namespace: string): Promise<an
...
+export async function waitForDeployment(name: string, namespace: string): Promise<void> {
+  console.log(`Trying to find deployment ${name} in namespace ${namespace}`);
+  for (let attempt = 0; attempt < 60; attempt++) {
+    try {
+      await exec(`./kubectl get deployment.apps/${name} -n ${namespace}`);
+    } catch (error) {
+      await sleep(1000);
+    }
+  }
+  console.log(`Found deployment ${name} in namespace ${namespace}`);
```
shaimendel: should we throw something in case we actually couldn't find the deployment? currently we print that anyway

test/setup/platforms/openshift4.ts
```diff
@@ -49,9 +49,35 @@ export async function exportKubeConfig(): Promise<void> {
...
+    kubectl.patchResourceFinalizers('customresourcedefinition', 'snykmonitors.charts.helm.k8s.io', 'snyk-monitor').catch((error) => {
+      console.log(error);
+    }),
+    kubectl.patchResourceFinalizers('snykmonitors.charts.helm.k8s.io', 'snyk-monitor', 'snyk-monitor').catch((error) => {
+      console.log(error);
+    }),
+  ]);
+
+  await Promise.all([
+    kubectl.deleteResource('customresourcedefinition', 'snykmonitors.charts.helm.k8s.io', 'default').catch((error) => {
```
shaimendel: I really don't like it, but I don't see a better alternative 😆

test/fixtures/operator/installation.yaml
ivanstanev: Haha, that is what OpenShift does under the hood when you click to install the Operator through the UI!

test/helpers/kubectl.ts
ivanstanev: Yes, but part of a general improvement, didn't want to touch too much code in this PR, only what is relevant

test/helpers/kubectl.ts
ivanstanev: It would be accurate indeed.. right now it would fail later, when it tries to wait for a deployment named `undefined` - it will fail immediately

test/setup/platforms/openshift4.ts
ivanstanev: How come? 😄 The Operator creates these files, we don't have a good way to clean them up...

test/helpers/kubectl.ts
shaimendel: capture somehow in jira please

test/setup/platforms/openshift4.ts
shaimendel: I mean that I don't like the `Promise.all([blabla.catch(err)])`

ivanstanev: Ah, fair enough... I find it cleaner than try-catching every individual promise... Plus the order doesn't matter (as long as they are grouped like that, so that `patch`-ing comes first)

PR #380, opened 2020-03-10
fix: upgrade K8s client to capture errors in Watch done() handlers

ivanstanev:

### What this does

Introduces a fix that allows errors to be emitted in Informer/Watch done() handlers.
This way we will be able to now see any API server errors, instead of receiving null in all cases.

### More information

- [The fix mentioned](https://github.com/kubernetes-client/javascript/commit/568a853dc10b41067f66e562853b2ca6d90d9c7b)
- [0.11.1 commits](https://github.com/kubernetes-client/javascript/commits/0.11.1)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -143,9 +143,9 @@
       }
     },
     "@kubernetes/client-node": {
-      "version": "0.10.3",
-      "resolved": "https://registry.npmjs.org/@kubernetes/client-node/-/client-node-0.10.3.tgz",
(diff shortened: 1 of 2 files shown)


ivanstanev:
Hmm, looking at the commit that they introduced in `0.11.1` I think we need to double-check how we treat the payload in the done() handler. It seems to be an array now?
Adding a DO NOT MERGE label for now

ivanstanev:
Reviewed the change and its PR (and the issue that PR links to), I think it is safe!
It will pass only a single error object, I think I got confused by the tests where I saw an array being used, but that's in a different context

PR #395, opened 2020-03-23
chore: remove CREATE_CLUSTER environment variable in tests

ivanstanev:
Make the platform setup entries generic enough so that they do not throw and do nothing whenever an action is not needed.
This allows us to execute actions in order and not be worried that they will throw, and allows us to remove the CREATE_CLUSTER environment variable. The platform itself will decide whether it needs to create a cluster, whether to delete or clean it, etc.

### More information

- [Jira ticket RUN-768](https://snyksec.atlassian.net/browse/RUN-768)
- #370 - addressing some of the comments


test/setup/index.ts
```diff
@@ -27,11 +26,9 @@ function getEnvVariableOrDefault(envVarName: string, defaultValue: string): stri
 export async function removeMonitor(): Promise<void> {
   await dumpLogs();
   try {
-    if (createCluster) {
-      await platforms[testPlatform].delete();
-    } else {
-      await platforms[testPlatform].clean();
-    }
+    await platforms[testPlatform].config();
```
shaimendel: why did you add the `config` call here?

test/setup/index.ts
```diff
@@ -90,17 +87,12 @@ export async function deployMonitor(): Promise<string> {
...
     await platforms[testPlatform].setupTester();
-    if (createCluster) {
-      await platforms[testPlatform].create();
-      await platforms[testPlatform].config();
-    } else {
-      await platforms[testPlatform].config();
-      await platforms[testPlatform].clean();
-    }
+    await platforms[testPlatform].create();
+    await platforms[testPlatform].config();
```
shaimendel: also here, just to make sure I understand - why did you remove the `clean` call here?

test/setup/index.ts
ivanstanev: Because of https://github.com/snyk/kubernetes-monitor/pull/395/files#diff-e5fa2adcbad65189e75788ac9a1fc12bR20-R28
At the start of the CI job/local test run, KUBECONFIG points to nothing

test/setup/index.ts
ivanstanev: Same as the previous comment, clean is now part of https://github.com/snyk/kubernetes-monitor/pull/395/files#diff-e5fa2adcbad65189e75788ac9a1fc12bR20-R28

PR #396, opened 2020-03-23
feat: docker-plugin with Distroless scanning support

Shesekino:

### What this does

bump docker-plugin with Distroless scanning support

### Notes for the reviewer

that's how I tested it locally:
1. build a local Distroless image
2. modify the system test to load it into the KinD cluster and run it
3. hack Skopeo's arguments during runtime so it doesn't try to save it from an image registry, but from my local docker daemon
4. observe the plugin scan the archive and return the same results as it does when running just the plugin on that image

I didn't enhance any of our tests since the logic of scanning Distroless belongs to the docker-plugin and not the Monitor. happy to enhance the testing in the Monitor if we think there's additional value in checking something specific.

### More information

https://snyksec.atlassian.net/browse/RUN-827

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3593,9 +3593,9 @@
       }
     },
     "snyk-docker-plugin": {
-      "version": "2.2.4",
-      "resolved": "https://registry.npmjs.org/snyk-docker-plugin/-/snyk-docker-plugin-2.2.4.tgz",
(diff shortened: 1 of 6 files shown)


shaimendel:
worths checking this again and see that we mention that the project is distroless in the UI

PR #399, opened 2020-03-24
chore: extract test tools into their own module

ivanstanev:
This allows to consolidate all the tools we use during tests into one place and make it easier to install them and clean them up.
This current change starts with skopeo and simplifies the system test, the idea is to do the same for all tools we use (kubectl, kind).

### DIFF
--- a/test/setup/tools/index.ts
+++ b/test/setup/tools/index.ts
@@ -1,0 +1,16 @@
+import {
+  path as skopeoPath,
+  install as installSkopeo,
+  remove as removeSkopeo,
+} from './skopeo';
+import { IThirdPartyTool } from './types';
+
+type SupportedTools = 'skopeo';
+
+export const tools: Record<SupportedTools, IThirdPartyTool> = {
+  skopeo: {
+    path: skopeoPath,
+    install: installSkopeo,
+    remove: removeSkopeo,
+  },
+};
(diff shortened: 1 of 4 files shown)


test/setup/tools/index.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  path as skopeoPath,
```
Shesekino: why not ```import * as skopeo from './skopeo'``` if you want the skopeo prefix?
your import will be 1 line long
and usage will be skopeo.path instead of skopeopath

test/setup/tools/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface IThirdPartyTool {
```
Shesekino: feels like over-abstracting things to be honest

test/setup/tools/skopeo.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!existsSync(directory)) {
+      mkdirSync(directory, { recursive: true });
+    }
+    await exec(`mv ./skopeo/skopeo ${path}`);
+    process.env['PATH'] = process.env['PATH'] + `:${directory}`;
+    await exec('rm -rf ./skopeo');
+  }
+}
+
+export async function remove() {
```
Shesekino: doesn't need to be async

ivanstanev:
Going to close this, I think you're right with the too many abstractions in tests!

PR #405, opened 2020-03-25
chore/cleanup tested operators

ivanstanev:

### What this does

The versioning scheme for our test Operators is `0.0.1-<commit sha>`.
Quay orders releases alphabetically when deciding what is the latest Operator release, meaning that the commit SHA with the highest string value wins - in the current case it is a several weeks old release starting with the SHA f6162b...
This means we have been testing the wrong release with every OpenShift test run.

The Operator catalog does not give us an easy way to pin to a specific version of the Operator, it will always choose the latest.
Hence as a workaround we can delete the Operator from Quay once we are done testing it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,6 +251,11 @@ jobs:
             docker pull ${KUBERNETES_MONITOR_IMAGE_NAME_AND_TAG} &&
             .circleci/do-exclusively --branch staging --job ${CIRCLE_JOB} npm run test:integration:openshift4
       - run:
+          name: Delete Operator from Quay
+          command: |
+            ./scripts/delete-operator-from-quay.sh
+          when: always
+      - run:
           name: Notify Slack on failure
           command: |
             if [[ "$CIRCLE_BRANCH" == "staging" ]]; then
--- a/scripts/delete-operator-from-quay.sh
+++ b/scripts/delete-operator-from-quay.sh
@@ -1,0 +1,13 @@
+#! /bin/bash
+
+set -e
+
+QUAY_TOKEN=$(curl -H "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d "{\"user\": {\"username\": \"${QUAY_USERNAME}\", \"password\": \"${QUAY_PASSWORD}\"}}" | jq -r .token)
+OPERATOR_VERSION="0.0.1-${CIRCLE_SHA1}"
+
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -250,6 +250,17 @@ jobs:
...
             docker pull ${KUBERNETES_MONITOR_IMAGE_NAME_AND_TAG} &&
             .circleci/do-exclusively --branch staging --job ${CIRCLE_JOB} npm run test:integration:openshift4
+      - run:
+          name: Delete Operator from Quay
+          command: |
+            export QUAY_TOKEN=$(curl -H "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d "{\"user\": {\"username\": \"${QUAY_USERNAME}\", \"password\": \"${QUAY_PASSWORD}\"}}" | jq -r .token)
+            export OPERATOR_VERSION="0.0.1-${CIRCLE_SHA1}"
+            STATUSCODE=$( curl --silent --output /dev/stderr --write-out "%{http_code}" -XDELETE -H "Accept: application/json" -H "Content-Type: application/json" -H "Authorization: $QUAY_TOKEN" "https://quay.io/cnr/api/v1/packages/snyk-runtime/snyk-operator/${OPERATOR_VERSION}/helm" )
+            if test $STATUSCODE -ge 300; then
+              echo "Unexpected status code $STATUSCODEE"
```
ivanstanev: Bug: should be STATUSCODE~E~

Shesekino: typo here? ```STATUSCODEE```

ivanstanev:
> the idea sounds good, but can we add less content to the circle ci config file please? it's already alarmingly huge

I am definitely up for that but did not want to spend more time on it since I already spent half a day  trying to work around the Operator Lifecycle Manager. In fact we have a Jira to track this debt but wasn't prioritised for this sprint: https://snyksec.atlassian.net/browse/RUN-839
WDYT about addressing it as part of that task, which will solve the growth of the config altogether?

Shesekino:
as I always say, Jira is where tasks go to die :) I'm concerned it will never be prioritised and never happen.
since we don't need to change the entire world, I'm only asking that we don't increase the existing debt.
how about starting with a small effort that could pave the way for bigger changes?

scripts/delete-operator-from-quay.sh
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
+
+set -e
+
+export QUAY_TOKEN=$(curl -H "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d "{\"user\": {\"username\": \"${QUAY_USERNAME}\", \"password\": \"${QUAY_PASSWORD}\"}}" | jq -r .token)
```
Shesekino: not sure you have to export those?

PR #416, opened 2020-03-30
test/openshift3 with yaml deployment

ivanstanev:

### What this does

Use a KinD cluster pinned to Kubernetes version 1.11.10. This allows us to simulate an OpenShift 3 environment.

### More information

- [Jira ticket RUN-826](https://snyksec.atlassian.net/browse/RUN-826)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 default_machine_config: &default_machine_config
   machine:
@@ -28,6 +28,20 @@ main_branches_filter: &main_branches_filter
       ignore:
         - staging
         - master
+
+commands:
+  setup_node12:
+    description: Setup Node 12
+    steps:
+      - run:
+          command: |
+            export NVM_DIR="/opt/circleci/.nvm"
+            [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+            nvm install v12
+            npm install
+            echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
+            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
+            echo 'nvm use 12' >> $BASH_ENV
 
 jobs:
 ######################## PR OR MERGE TO STAGING ########################
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -241,6 +241,26 @@ jobs:
       - store_artifacts:
           path: /tmp/logs/test/integration/eks
 
+  openshift3_integration_tests:
+    <<: *default_machine_config
+    steps:
+      - checkout
+      - setup_node12
+      - run:
+          name: Create temporary directory for logs
+          command: mkdir -p /tmp/logs/test/integration/openshift3
+      - run:
+          name: Integration tests OpenShift 3
+          command: |
+            export KUBERNETES_MONITOR_IMAGE_NAME_AND_TAG=$(./scripts/circleci-jobs/setup-openshift3-integration-tests.py)
+            npm run test:integration:openshift3
+      - run:
+          name: Notify Slack on failure
+          command: ./scripts/slack-notify-failure.sh "staging-openshift3-integration-tests-${CIRCLE_SHA1}"
+          when: on_fail
+      - store_artifacts:
+          path: /tmp/logs/test/integration/openshift3
+
   openshift4_integration_tests:
     <<: *default_machine_config
     steps:
@@ -434,6 +454,10 @@ workflows:
             - build_image
           <<: *staging_branch_only_filter
       - eks_integration_tests:
+          requires:
+            - build_image
+          <<: *staging_branch_only_filter
+      - openshift3_integration_tests:
           requires:
             - build_image
           <<: *staging_branch_only_filter
--- a/test/helpers/kubectl.ts
+++ b/test/helpers/kubectl.ts
@@ -1,21 +1,29 @@
 import { exec } from 'child-process-promise';
-import { chmodSync, writeFileSync, existsSync } from 'fs';
+import { chmodSync, writeFileSync, existsSync, unlinkSync } from 'fs';
 import { platform } from 'os';
 import { resolve } from 'path';
 import * as needle from 'needle';
 import * as sleep from 'sleep-promise';
 
-export async function downloadKubectl(): Promise<void> {
+/**
+ * @param version For example: "v1.18.0"
+ */
+export async function downloadKubectl(version: string): Promise<void> {
   const kubectlPath = resolve(process.cwd(), 'kubectl');
   if (existsSync(kubectlPath)) {
-    return;
+    if (version === 'latest') {
+      return;
+    }
+
+    // Always start clean when requesting a specific version.
+    unlinkSync(kubectlPath);
   }
 
-  console.log('Downloading kubectl...');
+  console.log(`Downloading kubectl ${version}...`);
 
   // eslint-disable-next-line @typescript-eslint/camelcase
   const requestOptions = { follow_max: 2 };
-  const k8sRelease = await getLatestStableK8sRelease();
+  const k8sRelease = version === 'latest' ? await getLatestStableK8sRelease() : version;
   const osDistro = platform();
   const bodyData = null;
   await needle('get', 'https://storage.googleapis.com/kubernetes-release/release/' +
--- a/test/helpers/kubectl.ts
+++ b/test/helpers/kubectl.ts
@@ -64,7 +64,7 @@ export async function applyK8sYaml(pathToYamlDeployment: string): Promise<void> 
 
 export async function createPodFromImage(name: string, image: string, namespace: string) {
   console.log(`Letting Kubernetes decide how to manage image ${image} with name ${name}`);
-  await exec(`./kubectl run ${name} --image=${image} -n ${namespace} -- sleep 999999999`);
+  await exec(`./kubectl run ${name} --generator=run-pod/v1 --image=${image} -n ${namespace} -- sleep 999999999`);
   console.log(`Done Letting Kubernetes decide how to manage image ${image} with name ${name}`);
 }
 
--- a/test/helpers/kubectl.ts
+++ b/test/helpers/kubectl.ts
@@ -1,32 +1,33 @@
 import { exec } from 'child-process-promise';
-import { accessSync, chmodSync, constants, writeFileSync } from 'fs';
+import { chmodSync, writeFileSync, existsSync } from 'fs';
 import { platform } from 'os';
 import { resolve } from 'path';
 import * as needle from 'needle';
 import * as sleep from 'sleep-promise';
 
 export async function downloadKubectl(): Promise<void> {
-  try {
-    accessSync(resolve(process.cwd(), 'kubectl'), constants.R_OK);
-  } catch (error) {
-    console.log('Downloading kubectl...');
-
-    // eslint-disable-next-line @typescript-eslint/camelcase
-    const requestOptions = { follow_max: 2 };
-    const k8sRelease = await getLatestStableK8sRelease();
-    const osDistro = platform();
-    const bodyData = null;
-    await needle('get', 'https://storage.googleapis.com/kubernetes-release/release/' +
-      `${k8sRelease}/bin/${osDistro}/amd64/kubectl`,
-      bodyData,
-      requestOptions,
-    ).then((response) => {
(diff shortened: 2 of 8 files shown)


test/helpers/kubectl.ts
```diff
@@ -1,21 +1,29 @@
...
 import { platform } from 'os';
 import { resolve } from 'path';
 import * as needle from 'needle';
 import * as sleep from 'sleep-promise';
 
-export async function downloadKubectl(): Promise<void> {
+/**
+ * @param version For example: "v1.18.0"
+ */
+export async function downloadKubectl(version?: string): Promise<void> {
```
Shesekino: I strongly prefer avoiding optional parameters if possible

test/helpers/kubectl.ts
```diff
@@ -1,21 +1,29 @@
...
+    if (version === undefined) {
+      return;
+    }
+
+    // Always start clean when requesting a specific version.
+    unlinkSync(kubectlPath);
   }
 
-  console.log('Downloading kubectl...');
+  console.log(`Downloading kubectl ${version || ''}...`);
```
Shesekino: not sure about it, it looks like code written for backwards compatibility instead of a function written in the first place to solve the problem of "download Kubectl"

test/helpers/kubectl.ts
```diff
@@ -72,7 +72,7 @@ export async function applyK8sYaml(pathToYamlDeployment: string): Promise<void>
 
 export async function createPodFromImage(name: string, image: string, namespace: string) {
   console.log(`Letting Kubernetes decide how to manage image ${image} with name ${name}`);
-  await exec(`./kubectl run ${name} --image=${image} -n ${namespace} -- sleep 999999999`);
+  await exec(`./kubectl run ${name} --generator=run-pod/v1 --image=${image} -n ${namespace} -- sleep 999999999`);
```
Shesekino: so this is backwards compatibility for the older version of kubectl the OS3 tests will use, correct? where the default generator was deployment rather than pod?
I wonder if it's worth mentioning more explicitly somehow, but up to you

test/setup/index.ts
```diff
@@ -92,10 +92,13 @@ export async function deployMonitor(): Promise<string> {
 
     console.log(`platform chosen is ${testPlatform}, createCluster===${createCluster}`);
 
-    await kubectl.downloadKubectl();
+    // TODO: hack, rewrite this
+    const kubernetesVersion = testPlatform === 'openshift3' ? 'v1.11.10' : undefined;
```
Shesekino: agreed, I would want this to come from the module responsible for Openshift/Openshift3

test/setup/index.ts
```diff
@@ -92,10 +92,13 @@ export async function deployMonitor(): Promise<string> {
...
+    await kubectl.downloadKubectl(kubernetesVersion);
```
Shesekino: if we let each platform decide (or not decide) which kubectl version it wants, maybe indicated by "latest" instead of undefined, we would be able to avoid some of the code written to handle the undefined parts, wdyt?
we'd still need to handle "latest" of course but I think it could be clearer

test/setup/platforms/index.ts
```diff
@@ -35,6 +35,16 @@ const eksSetup: IPlatformSetup = {
   setupTester: eks.setupTester,
 };
 
+// Use a kind cluster pinned to a specific Kubernetes version to mimic OS3.
```
Shesekino: why is that?
I mean, I believe you 😛 but maybe there's a doc somewhere about it? how did you learn this?
might be worth documenting in the code around here

test/setup/platforms/kind.ts
```diff
@@ -11,9 +11,9 @@ export async function setupTester(): Promise<void> {
   await download(osDistro);
 }
 
-export async function createCluster(): Promise<void> {
+export async function createCluster(version?: string): Promise<void> {
```
Shesekino: same idea about the ```?```

scripts/circleci-jobs/openshift3-integration-tests.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python3
+
+import os
+import subprocess
+
+DOCKERHUB_USER = os.getenv('DOCKERHUB_USER')
+DOCKERHUB_PASSWORD = os.getenv('DOCKERHUB_PASSWORD')
+PROCESS = subprocess.run(["docker", "login", "--username", DOCKERHUB_USER, "--password", DOCKERHUB_PASSWORD], capture_output=True, text=True)
```
Shesekino: I think you can use ```subprocess.get_output``` for that

scripts/circleci-jobs/openshift3-integration-tests.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python3
+
+import os
+import subprocess
+
+DOCKERHUB_USER = os.getenv('DOCKERHUB_USER')
```
Shesekino: I would use capital letters for maybe some of the consts you want to shout about, but not for all vars. wdyt? makes it a bit hard to read

scripts/circleci-jobs/openshift3-integration-tests.py
```diff
@@ -0,0 +1,23 @@
...
+DOCKERHUB_PASSWORD = os.getenv('DOCKERHUB_PASSWORD')
+PROCESS = subprocess.run(["docker", "login", "--username", DOCKERHUB_USER, "--password", DOCKERHUB_PASSWORD], capture_output=True, text=True)
+print(PROCESS.stdout)
+
+CIRCLE_BRANCH = os.getenv("CIRCLE_BRANCH")
+if CIRCLE_BRANCH == "staging":
+    IMAGE_TAG = "staging-candidate"
+else:
+    IMAGE_TAG = "discardable"
```
Shesekino: we're running os3 integration tests for regular branches as well? not just staging? I don't think we do that for all non-KinD integration tests, not sure we want to start doing it

scripts/circleci-jobs/openshift3-integration-tests.py
```diff
@@ -0,0 +1,23 @@
...
+CIRCLE_SHA1 = os.getenv("CIRCLE_SHA1")
+KUBERNETES_MONITOR_IMAGE_NAME_AND_TAG = "snyk/kubernetes-monitor:" + IMAGE_TAG + "-" + CIRCLE_SHA1
+
+PROCESS = subprocess.run(["docker", "pull", KUBERNETES_MONITOR_IMAGE_NAME_AND_TAG], capture_output=True, text=True)
+print(PROCESS.stdout)
+
+PROCESS = subprocess.run(["npm", "run", "test:integration:openshift3"], capture_output=True, text=True, env=dict(os.environ))
```
Shesekino: it looks like this script does two things:
1. sets up the integration test environment by pulling the right image
2. wraps npm test

why not let it just handle the setup, and let the circleCI call npm test as it does for all other flows?

scripts/circleci-jobs/openshift3-integration-tests.py
ivanstanev: Agreed but this is a bit of a KinD-based integration test which allows us to run it on PRs. The other platform test suffer from requiring exclusivity on a single cluster, which would slow down PRs if opened in parallel.

test/setup/platforms/index.ts
ivanstanev: I think we decided not to try spinning an OpenShift cluster but instead use a regular Kubernetes cluster with the same version that OS3 ships. So that would be 1.11:  https://docs.openshift.com/container-platform/3.11/release_notes/ocp_3_11_release_notes.html#ocp-311-about-this-release
> This release is based on OKD 3.11, and it uses Kubernetes 1.11.

test/setup/index.ts
ivanstanev: I find that I'll have to put another variable in the IPlatform interface to achieve this, do we really want to grow that? Or write some function `getKubernetesVersionForPlatform(testPlatform: string)` in the index.ts of the platforms module which will do a switch-case/if for all platforms?

Shesekino: your call, I just really don't want ```test/setup/index.ts``` to have more and more references to each specific platform. one of the reasons we reverted the original Openshift4 PR was just that.
I really want a clean flow of events with clean abstraction as much as possible.
happy to 🍐 on that if you want

test/setup/platforms/index.ts
Shesekino: not sure I was party to this decision, but all the same - why are we satisfied with that?
that is, when does using KinD give us enough confidence to claim we support OS3?

scripts/circleci-jobs/openshift3-integration-tests.py
Shesekino: hmmm okay, let's start with your proposal, but be open minded towards pushing it to run only on merges if at some point we decide it's too much hassle to run on every branch?

test/setup/platforms/index.ts
ivanstanev: Because OS3 is just a bunch of components built on top of vanilla Kubernetes. More specifically, it's lots of Deployments and CustomResources utilising the Kubernetes API. There aren't any modifications to the core Kubernetes underneath it. So any potential differences would happen at runtime (e.g. some component may decide to block insecure Deployments) and are not at the core platform.

We don't do anything special with our Deployment, and we have verified it on both vanilla Kubernetes and on OpenShift 4. Both of these give us already a fair amount of confidence that it would just work on OS3 as well, because OS3 and OS4 are not very different in terms of behaviour. Also OS3 support is mostly maintenance updates/patches so we really don't expect something major changing in the platform!

Shesekino: excellent, thanks for the explanation :)

scripts/circleci-jobs/openshift3-integration-tests.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/python3
+
+import os
```
shaimendel: what is specific here to openshift3? wondering about the file name and why it is used only in this flow

.circleci/config.yml
```diff
@@ -29,6 +29,20 @@ main_branches_filter: &main_branches_filter
...
+  setup_node12:
+    description: Setup Node 12
+    steps:
+      - run:
+          command: |
+            export NVM_DIR="/opt/circleci/.nvm"
+            [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+            nvm install v12
+            npm install
+            echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
```
shaimendel: this is instead of aliasing v12 to default?

scripts/circleci-jobs/openshift3-integration-tests.py
ivanstanev: Nothing specific, the goal is to reduce our CCI script size. Starting with this PR by not adding more to the existing debt

.circleci/config.yml
ivanstanev: Not sure what you mean

scripts/circleci-jobs/openshift3-integration-tests.py
shaimendel: so maybe it makes sense to call it in a more general name that hints about the file purpose? anyway, your call

.circleci/config.yml
shaimendel: me too

shaimendel: I thought that `nvm alias default 12` might do a similar work

test/setup/platforms/index.ts
```diff
@@ -44,6 +44,10 @@ const openshift4Setup: IPlatformSetup = {
   setupTester: openshift4.setupTester,
 };
 
+export function getKubernetesVersionForPlatform(testPlatform: string): string {
```
Shesekino: I'm missing something, why are we accepting ```testPlatform``` but not using it?

test/setup/platforms/index.ts
```diff
@@ -45,11 +55,17 @@ const openshift4Setup: IPlatformSetup = {
 };
 
 export function getKubernetesVersionForPlatform(testPlatform: string): string {
-  return 'latest';
+  switch (testPlatform) {
```
Shesekino: ah, it was put in separate commits 😬 found my answer on the next commit

scripts/circleci-jobs/openshift3-integration-tests.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/python3
```
Shesekino: maybe rename the file to ```setup-openshift3-integration-test``` or something like that?

PR #418, opened 2020-03-31
fix: remove runAsUser/Group to support OpenShift 3

ivanstanev:
These values are automatically generated by OpenShift and cannot be the ones we have been enforcing until now.


snyk-monitor-deployment.yaml
```diff
@@ -66,8 +66,6 @@ spec:
             cpu: '1'
             memory: '2Gi'
         securityContext:
-          runAsUser: 10001
```
shaimendel: is it "ok" to delete it also from here and the helm chart? what does it mean that we don't have it?

ivanstanev: It runs as the user and group specified in the container image (which are the same: https://github.com/snyk/kubernetes-monitor/blob/staging/Dockerfile#L36-L40) so we're good!

PR #427, opened 2020-04-02
chore: reorganised scripts folder

agatakrajewska:

### What this does

As a perquisite for https://snyksec.atlassian.net/browse/RUN-828 & https://snyksec.atlassian.net/browse/RUN-828 - building kubernetes-monitor with RedHat UBI, we need to organise the growing ./circleci/config.yaml and our already existing bash scripts. 

### More information

- [Jira ticket RUN-840](https://snyksec.atlassian.net/browse/RUN-840)


ivanstanev:
Could we also remove the `slack-` prefix on all scripts under the `slack/` directory?

scripts/slack/notify-failure-on-branch.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+NOTIFY_BRANCH_NAME="staging"
+DISPLAY_BRANCH_NAME="$1"
+
+if [[ "$CIRCLE_BRANCH" == "$NOTIFY_BRANCH_NAME" ]]; then
+  ./scripts/slack/notify-failure.sh "$DISPLAY_BRANCH_NAME"
+else
+  echo "Current branch is $CIRCLE_BRANCH so skipping notifying Slack"
+fi
```
ivanstanev, line 10: Missing end line here

scripts/slack/notify-failure-on-branch.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+NOTIFY_BRANCH_NAME="staging"
+DISPLAY_BRANCH_NAME="$1"
+
+if [[ "$CIRCLE_BRANCH" == "$NOTIFY_BRANCH_NAME" ]]; then
```
ivanstanev, line 6: Wondering if we could directly compare instead of using a variable
```suggestion
if [[ "$CIRCLE_BRANCH" == "staging" ]]; then
```

agatakrajewska: we totally can, I've put it as a variable so if we want to parametrise it, it will be super easy, but I guess it will only always be staging?

ivanstanev: I think the script is really tiny and adding an extra variable makes it harder to read because now you have to go back and figure out what value it is assigned. A variable would make more sense if the pattern started repeating 🙏Just a suggestion, up to you if you want to change it!

agatakrajewska: that makes sense, let's add it when we need it! 👍

PR #429, opened 2020-04-03
feat: build snyk-monitor with a Red Hat UBI 8 base image

ivanstanev:

### What this does

As part of the Red Hat certification process we need to use a Red Hat approved base image for our product.

### More information

- [Jira ticket RUN-828](https://snyksec.atlassian.net/browse/RUN-828)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,18 +12,20 @@ RUN cd $GOPATH/src/github.com/containers/skopeo \
 #---------------------------------------------------------------------
 # STAGE 2: Build the kubernetes-monitor
 #---------------------------------------------------------------------
-FROM node:erbium-alpine
+FROM registry.access.redhat.com/ubi8/ubi:latest
 
 LABEL maintainer="Snyk Ltd"
 
 ENV NODE_ENV production
 
-RUN apk update
-RUN apk upgrade
-RUN apk --no-cache add dumb-init
+RUN curl -sL https://rpm.nodesource.com/setup_12.x | bash -
+RUN yum install -y nodejs
 
-RUN addgroup -S -g 10001 snyk
-RUN adduser -S -G snyk -h /srv/app -u 10001 snyk
+RUN curl -L -o /usr/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64
+RUN chmod 755 /usr/bin/dumb-init
+
+RUN groupadd -g 10001 snyk
+RUN useradd -g snyk -d /srv/app -u 10001 snyk
 
 WORKDIR /srv/app
 USER 10001:10001
@@ -44,6 +46,9 @@ RUN npm install
 
 # add the rest of the app files
 ADD --chown=snyk:snyk . .
+
+# OpenShift 4 doesn't allow dumb-init access the app folder without this permission.
+RUN chmod 755 /srv/app && chmod 755 /srv/app/bin && chmod +x /srv/app/bin/start
 
 # Complete any `prepare` tasks (e.g. typescript), as this step ran automatically prior to app being copied
 RUN npm run prepare


.circleci/config.yml
```diff
@@ -52,16 +52,23 @@ commands:
...
-            IMAGE_NAME_CANDIDATE=snyk/kubernetes-monitor:${IMAGE_TAG}-${CIRCLE_SHA1} &&
-            ./scripts/docker/build-image.sh ${IMAGE_NAME_CANDIDATE} &&
+            docker login --username "${CONTAINER_REGISTRY_USERNAME}" --password "${CONTAINER_REGISTRY_PASSWORD}" registry.redhat.io
+            IMAGE_TAG=$([[ "$CIRCLE_BRANCH" == "staging" ]] && echo "staging-candidate" || echo "discardable")
+            IMAGE_NAME_CANDIDATE=snyk/kubernetes-monitor:${IMAGE_TAG}-${CIRCLE_SHA1}
+            RED_HAT_AUTH_CREDENTIALS_FILE=credentials
+            echo "${RED_HAT_USERNAME}" >> ${RED_HAT_AUTH_CREDENTIALS_FILE}
+            echo "${RED_HAT_PASSWORD}" >> ${RED_HAT_AUTH_CREDENTIALS_FILE}
+            # TODO: only Docker version 19.03.3 with BuildKit can pull the image using the FROM directive in the Dockerfile
+            docker pull registry.access.redhat.com/ubi8/ubi:latest
```
shaimendel, line 69: are we ok with always using `latest` vs using a specific version?

.circleci/config.yml
```diff
@@ -52,16 +52,23 @@ commands:
...
           name: Build image
           command: |
-            docker login --username ${DOCKERHUB_USER} --password ${DOCKERHUB_PASSWORD} &&
-            export IMAGE_TAG=$([[ "$CIRCLE_BRANCH" == "staging" ]] && echo "staging-candidate" || echo "discardable") &&
-            IMAGE_NAME_CANDIDATE=snyk/kubernetes-monitor:${IMAGE_TAG}-${CIRCLE_SHA1} &&
-            ./scripts/docker/build-image.sh ${IMAGE_NAME_CANDIDATE} &&
+            docker login --username "${CONTAINER_REGISTRY_USERNAME}" --password "${CONTAINER_REGISTRY_PASSWORD}" registry.redhat.io
+            IMAGE_TAG=$([[ "$CIRCLE_BRANCH" == "staging" ]] && echo "staging-candidate" || echo "discardable")
+            IMAGE_NAME_CANDIDATE=snyk/kubernetes-monitor:${IMAGE_TAG}-${CIRCLE_SHA1}
+            RED_HAT_AUTH_CREDENTIALS_FILE=credentials
```
shaimendel, line 65: can we push this image to a snyk org and use it from there? so we do this manually one time, and the local build and CI/CD stays basically the same

Dockerfile
```diff
@@ -1,37 +1,23 @@
...
 
-RUN apk update
-RUN apk upgrade
-RUN apk --no-cache add dumb-init
+RUN curl -sL https://rpm.nodesource.com/setup_12.x | bash -
+RUN yum install -y nodejs
 
-RUN addgroup -S -g 10001 snyk
-RUN adduser -S -G snyk -h /srv/app -u 10001 snyk
+RUN --mount=type=secret,id=credentials cat /run/secrets/credentials | subscription-manager register --auto-attach
```
shaimendel, line 11: why can't we build skopeo in the old manner?

.circleci/config.yml
ivanstanev: I think so, I assume we would be getting mostly security fixes with new releases and no breaking changes

.circleci/config.yml
ivanstanev: I think it's not relevant any more

Dockerfile
```diff
@@ -43,4 +42,4 @@ ADD --chown=snyk:snyk . .
 # Complete any `prepare` tasks (e.g. typescript), as this step ran automatically prior to app being copied
 RUN npm run prepare
 
-ENTRYPOINT ["/usr/bin/dumb-init", "--", "bin/start"]
+ENTRYPOINT ["bin/start"]
```
Shesekino, line 45: what about dumb-init?  does the UBI have something similar?

ivanstanev: Yes, UBI comes with its own init manager 🙏 
`ubi-minimal` doesn't so that's why I opted for the full `ubi`; would've been nice to have the slimmer image

Dockerfile
```diff
@@ -12,18 +12,17 @@ RUN cd $GOPATH/src/github.com/containers/skopeo \
...
 # STAGE 2: Build the kubernetes-monitor
 #---------------------------------------------------------------------
-FROM node:erbium-alpine
+FROM registry.access.redhat.com/ubi8/ubi:latest
 
 LABEL maintainer="Snyk Ltd"
 
 ENV NODE_ENV production
 
-RUN apk update
```
ivanstanev, line 21: Should I preserve the alternative: `yum update` before doing anything else?

Dockerfile
```diff
@@ -12,18 +12,17 @@ RUN cd $GOPATH/src/github.com/containers/skopeo \
...
-RUN apk upgrade
-RUN apk --no-cache add dumb-init
+RUN curl -sL https://rpm.nodesource.com/setup_12.x | bash -
+RUN yum install -y nodejs
```
shaimendel, line 22: can't we somehow install dumb-init?

ivanstanev: dumb-init is a drop-in for when an init system doesn't exist, `ubi` has one already:
https://github.com/snyk/kubernetes-monitor/pull/429#discussion_r403958309

Dockerfile
shaimendel: does UBI's init manager has the same capabilities of dumb-init? in terms of proxying signals etc

Dockerfile
```diff
@@ -47,6 +47,9 @@ RUN npm install
 # add the rest of the app files
 ADD --chown=snyk:snyk . .
 
+# OpenShift 4 doesn't allow the user can't access the app files without giving permission access.
+RUN chmod +x .
```
shaimendel, line 51: 1. if possible to improve the syntax of the comment :) 
2. why do you need the +x? what is it trying to execute? I wonder if +r or +w should be enough
3. don't you need to make that recursive? I find it weird that it works 😆

shaimendel: pay attention that your chmod +x patch only applies to the app folder, I wonder if for some reason chmod +x bin will be enough

agranado2k: The problem was that `dumb-init` was trying to access the `bin/start` and it didn't have permission. I tried to be more specific to access only `bin/start` and it didn't work. But let me try your suggestions, I prefer it would be more specific than generic too.

PR #430, opened 2020-04-03
chore: slack integration scripts rewritten to python

agatakrajewska:

### What this does

We wanted to clean up circleci scripts, rewriting them into python so they are environment agnostic.

### More information

- [Jira ticket RUN-840](https://snyksec.atlassian.net/browse/RUN-840)

### Screenshots

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,14 @@ commands:
             echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
             echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
             echo 'nvm use 12' >> $BASH_ENV
+  install_python_requests:
+    description: Install requests library
+    steps:
+      - run:
+          command: |
+            sudo apt update
+            sudo apt install python3-requests
+           
 
 jobs:
 ######################## PR OR MERGE TO STAGING ########################
@@ -55,6 +63,7 @@ jobs:
     <<: *default_machine_config
     steps:
       - checkout
+      - install_python_requests
       - run:
           name: Build image
           command: |
@@ -66,13 +75,14 @@ jobs:
       - run:
           name: Notify Slack on failure
           command: |
-            ./scripts/slack/notify-failure-on-branch.sh "staging-build-image-${CIRCLE_SHA1}"
+            ./scripts/slack/notify_failure_on_branch.py "staging-build-image-${CIRCLE_SHA1}"
           when: on_fail
 
   build_operator:
     <<: *default_machine_config
     steps:
       - checkout
+      - install_python_requests
       - run:
           name: Download Operator SDK
           command: |
@@ -105,7 +115,7 @@ jobs:
       - run:
           name: Notify Slack on failure
           command: |
-            ./scripts/slack/notify-failure-on-branch.sh "staging-build-operator-${CIRCLE_SHA1}"
+            ./scripts/slack/notify_failure_on_branch.py "staging-build-operator-${CIRCLE_SHA1}"
           when: on_fail
 
   upload_operator:
@@ -131,6 +141,7 @@ jobs:
     steps:
       - checkout
       - setup_node12
+      - install_python_requests
       - run:
           name: Unit tests
           command: |
@@ -140,7 +151,7 @@ jobs:
       - run:
           name: Notify Slack on failure
           command: |
-            ./scripts/slack/notify-failure-on-branch.sh "staging-unit-tests-${CIRCLE_SHA1}"
+            ./scripts/slack/notify_failure_on_branch.py "staging-unit-tests-${CIRCLE_SHA1}"
           when: on_fail
   
   system_tests:
@@ -148,6 +159,7 @@ jobs:
     steps:
       - checkout
       - setup_node12
+      - install_python_requests
       - run:
           name: System tests
           command: |
@@ -156,7 +168,7 @@ jobs:
       - run:
           name: Notify Slack on failure
           command: |
-            ./scripts/slack/notify-failure-on-branch.sh "staging-system-tests-${CIRCLE_SHA1}"
+            ./scripts/slack/notify_failure_on_branch.py "staging-system-tests-${CIRCLE_SHA1}"
           when: on_fail
 
   integration_tests:
@@ -164,6 +176,7 @@ jobs:
     steps:
       - checkout
       - setup_node12
+      - install_python_requests
       - run:
           name: create temp dir for logs
           command: mkdir -p /tmp/logs/test/integration/kind
@@ -175,7 +188,7 @@ jobs:
       - run:
           name: Notify Slack on failure
           command: |
-            ./scripts/slack/notify-failure-on-branch.sh "staging-integration-tests-${CIRCLE_SHA1}"
+            ./scripts/slack/notify_failure_on_branch.py "staging-integration-tests-${CIRCLE_SHA1}"
           when: on_fail
       - store_artifacts:
           path: /tmp/logs/test/integration/kind
@@ -185,6 +198,7 @@ jobs:
     steps:
       - checkout
       - setup_node12
+      - install_python_requests
       - run:
           name: Create temporary directory for logs
           command: mkdir -p /tmp/logs/test/integration/kind-helm
@@ -196,7 +210,7 @@ jobs:
       - run:
           name: Notify Slack on failure
           command: |
-            ./scripts/slack/notify-failure-on-branch.sh "staging-integration-helm-tests-${CIRCLE_SHA1}"
+            ./scripts/slack/notify_failure_on_branch.py "staging-integration-helm-tests-${CIRCLE_SHA1}"
           when: on_fail
       - store_artifacts:
           path: /tmp/logs/test/integration/kind-helm
@@ -205,6 +219,7 @@ jobs:
     <<: *default_machine_config
     steps:
       - checkout
+      - install_python_requests
       - run:
           name: Create temp dir for logs
           command: mkdir -p /tmp/logs/test/integration/eks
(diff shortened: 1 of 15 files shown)


.circleci/config.yml
```diff
@@ -51,6 +51,10 @@ commands:
 
 jobs:
 ######################## PR OR MERGE TO STAGING ########################
+  check_python_version:
```
agatakrajewska, line 54: this job is here only for testing, will remove

.circleci/config.yml
```diff
@@ -48,13 +48,22 @@ commands:
             echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
             echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
             echo 'nvm use 12' >> $BASH_ENV
+  pip_install_requests:
+    description: Install requests library
+    steps:
+      - run:
+          command: |
+            sudo apt update
+            sudo apt install python3-requests
```
Shesekino, line 57: do we not have ```pip``` there?

scripts/slack/notify_deploy.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python3
+
+import sys
+import requests
+import json
+
+image_name = sys.argv[2]
```
Shesekino, line 7: what's in argv[1]?

scripts/slack/notify_failure.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/python3
+
+import sys
+import requests
+import json
+import os
+
+def notifyslack( branch_name ):
```
Shesekino, line 8: notifySlack

Shesekino, line 8: no need for spaces between brackets

Shesekino, line 8: I think it might be nicer if all the scripts are consistent in exposing a function, even if right now some of them are only called directly

scripts/slack/notify_failure_on_branch.py
```diff
@@ -0,0 +1,14 @@
...
+import subprocess
+from notify_failure import notifyslack
+
+display_branch_name = sys.argv[1]
+circle_branch_name = os.getenv('CIRCLE_BRANCH')
+
+if circle_branch_name == 'staging':
+    notifyslack(display_branch_name)
+else:
+    print('Current branch is ' + circle_branch_name + 'so skipping notifying Slack')
```
Shesekino, line 14: missing space after the 2nd plus?

scripts/slack/notify_push.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python3
+
+import requests
+import json
+
+def notifyslack( branch_name ):
```
Shesekino, line 6: notifySlack, no spaces between brackets

scripts/slack/notify_push.py
```diff
@@ -0,0 +1,27 @@
...
+          'color': '#7CD197',
+          'fallback': 'Build Notification: ' + circle_build_url,
+          'title': 'Kubernetes-Monitor Publish Notification',
+          'text': ':egg_fancy: Published Kubernetes-Monitor: `' + branch_name + '` :egg_fancy:'
+        }
+      ]
+    }
+
+    requests.post(url, data = json.dumps(data))
+    return
```
Shesekino, line 23: not sure we need the ```return``` here

scripts/slack/notify_success_no_publish.py
```diff
@@ -0,0 +1,21 @@
...
+    {
+      'color': '#7CD197',
+      'fallback': 'Build Notification: ' + circle_build_url,
+      'title': 'Kubernetes-Monitor Publish Notification',
+      'text': ':egg_fancy: Successful `master` merge, but no `gh-pages` release occurring :egg_fancy:'
+    }
+  ]
+}
+
+requests.post(url, data = json.dumps(data))
```
Shesekino, line 21: can remove the spaces around the ```=```
(in other places as well, when passed as named arguments to a function)

scripts/slack/notify_success_no_release.py
```diff
@@ -0,0 +1,21 @@
...
+    {
+      'color': '#7CD197',
+      'fallback': 'Build Notification: ' + circle_build_url,
+      'title': 'Kubernetes-Monitor Publish Notification',
+      'text': ':egg_fancy: Successful `staging` merge, but no semantic-release occurring :egg_fancy:'
+    }
+  ]
+}
+
+requests.post(url, data = json.dumps(data))
```
Shesekino, line 21: are you sure we need the ```json.dumps``` here? and in the other places
https://www.w3schools.com/python/ref_requests_post.asp

.circleci/config.yml
agatakrajewska: There is some sort of ssl problem using requests installed with pip and python2. That  problem is fixed in python3 so to be able to use requests with python3 this was the way it had to be installed after lots of trial end error. Hope it makes sense? This is the slack thread with all the detail: https://snyk.slack.com/archives/CDSMEJ29E/p1586268718117900

ivanstanev: The name of the job is probably misleading though: `pip_install_requests` but doesn't use pip to install! Maybe a more generic `install_requests_python`. Or alternatively the very specific `apt_install_python3_requests` 🤷‍♂ Up to you!

scripts/slack/notify_deploy.py
agatakrajewska: that should definitely be 1 and 2 index below, fixing now.

scripts/slack/notify_failure.py
agatakrajewska: Good point 👍

scripts/slack/notify_deploy.py
ivanstanev: If we look at the usage of this notify_deploy script:
```sh
./scripts/slack/notify_deploy.py $LATEST_TAG dev
```
I think argv[1] would be the name of the tag, and argv[2] the deployment env... so argv[3] doesn't exist, right? I think the indices are 1 too many

scripts/slack/notify_failure.py
```diff
@@ -0,0 +1,29 @@
...
+          'color': '#EE0000',
+          'fallback': 'Build Notification: ' + circle_build_url,
+          'title': ':warning: Kubernetes-Monitor Merge Failure :warning:',
+          'text': ':egg_broken_1: Kubernetes-Monitor broken branch: `' + branch_name + '` :egg_broken_1:\n' + circle_build_url
+        }
+      ]
+    }
+
+    requests.post(url, data = json.dumps(data))
+    return
```
ivanstanev, line 25: Also not sure we need a return here

scripts/slack/notify_push.py
agatakrajewska: For some reason I was under the impression that every method in python needs a return statement! I just double checked and it does not, if it's skipped 'None' is returned 👍

scripts/slack/notify_success_no_release.py
agatakrajewska: Hmm maybe not! Looks like I can send it a json parameter, looking into it now

.circleci/config.yml
agatakrajewska: you're right! I've changed how I install requests, but didn't change the name of the job 👍

scripts/slack/notify_deploy.py
agatakrajewska: that is fixed now!

scripts/slack/notify_deploy.py
```diff
@@ -0,0 +1,27 @@
...
+    url = os.getenv('SLACK_WEBHOOK')
+
+    data = {
+      'attachments': 
+      [
+        {
+          'color': '#7CD197',
+          'fallback': 'Build Notification: ' + circle_build_url,
+          'title': 'Kubernetes-Monitor Deploy Notification',
+          'text': ':hatching_chick: Deploying Kubernetes-Monitor on `'+ deployment_env_name + '`:`' + image_name + '`: :hatching_chick:'
```
ivanstanev, line 17: I think it needs a space after the colon (if you look at the shell script it had a space):
```python
'`: `'
```
There also isn't a colon before the chick emoji:
```python
` :hatching_chick:'
```

scripts/slack/notify_success_no_publish.py
```diff
@@ -0,0 +1,24 @@
...
+          'title': 'Kubernetes-Monitor Publish Notification',
+          'text': ':egg_fancy: Successful `master` merge, but no `gh-pages` release occurring :egg_fancy:'
+        }
+      ]
+    }
+
+    requests.post(url, json=data)
+
+if __name__ == '__main__':
+    notifySlack()
```
ivanstanev, line 24: Missing endline 🙏

scripts/slack/notify_success_no_release.py
```diff
@@ -0,0 +1,24 @@
...
+          'title': 'Kubernetes-Monitor Publish Notification',
+          'text': ':egg_fancy: Successful `staging` merge, but no semantic-release occurring :egg_fancy:'
+        }
+      ]
+    }
+
+    requests.post(url, json=data)
+
+if __name__ == '__main__':
+    notifySlack()
```
ivanstanev, line 24: Missing endline 🙏

scripts/slack/notify_deploy.py
ivanstanev: The space after the colon is missing though!
Currently:
```python
deployment_env_name + '`:`' + image_name
```
How it should be:
```python
deployment_env_name + '`: `' + image_name
```

agatakrajewska: yes, sorry, I've added it in the wrong place 🤦‍♀

scripts/slack/notify_push.py
Shesekino: yes, just like ```undefined``` for Node :)

PR #433, opened 2020-04-08
fix: passing data keyword, instead of json, to requests.post()

agatakrajewska:

### What this does

For some reason passing `json` keyword into python `requests.post()` fails on some container configs, I'm expecting different python versions (?). As a solution, I'm passing `data` and parsing a dictionary into json with a lib.

### DIFF
--- a/scripts/slack/notify_deploy.py
+++ b/scripts/slack/notify_deploy.py
@@ -2,6 +2,7 @@
 
 import os
 import requests
+import json
 import sys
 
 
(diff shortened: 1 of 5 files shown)


agatakrajewska:
> Looks good!
> 
> In future changes it would be better to include a short description why we're changing this. Today we know why, but 6 months from now we probably will have forgotten and someone may attempt to do the same thing and wonder why it's failing

You mean description in the commit? Noted!

PR #443, opened 2020-04-17
Feat/proxying

ivanstanev:

### What this does

Add proxying support (for forward proxying).
The proxying intentionally excludes requests to the Kubernetes API server - those will never be proxied.

When using NO_PROXY, we do an exact match on the hostname with the provided list of hosts. So CIDR ranges (192.168.0.0/16) and wildcards (*.example.domain) are not supported.

### Notes for the reviewer

Review commit by commit as usual 🙏 

### More information

- [Jira ticket RUN-875](https://snyksec.atlassian.net/browse/RUN-875)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "test:system": "tap test/system --timeout=600",
     "test:integration:kind:yaml": "DEPLOYMENT_TYPE=YAML TEST_PLATFORM=kind CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
     "test:integration:kind:helm": "DEPLOYMENT_TYPE=Helm TEST_PLATFORM=kind CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
+    "test:integration:kind:proxy": "DEPLOYMENT_TYPE=Proxy TEST_PLATFORM=kind CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
     "test:integration:eks:yaml": "DEPLOYMENT_TYPE=YAML TEST_PLATFORM=eks CREATE_CLUSTER=false tap test/integration/kubernetes.test.ts --timeout=900",
     "test:integration:openshift3:yaml": "DEPLOYMENT_TYPE=YAML TEST_PLATFORM=openshift3 CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
     "test:integration:openshift4:operator": "DEPLOYMENT_TYPE=Operator TEST_PLATFORM=openshift4 CREATE_CLUSTER=false tap test/integration/kubernetes.test.ts --timeout=900",
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@types/needle": "^2.0.4",
     "@types/node": "^10.17.17",
     "@types/sinon": "^7.5.2",
+    "@types/tunnel": "0.0.1",
     "async": "^2.6.2",
     "aws-sdk": "^2.633.0",
     "bunyan": "^1.8.12",
@@ -42,6 +43,7 @@
     "snyk-config": "3.0.0",
     "snyk-docker-plugin": "2.8.0",
     "source-map-support": "^0.5.16",
+    "tunnel": "0.0.6",
     "typescript": "^3.8.3",
     "ws": "^7.2.1",
     "yaml": "^1.7.2"
--- a/snyk-monitor-cluster-permissions.yaml
+++ b/snyk-monitor-cluster-permissions.yaml
@@ -87,4 +87,7 @@ data:
   namespace: ""
   integrationApi: ""
   clusterName: ""
+  http_proxy: ""
+  https_proxy: ""
+  no_proxy: ""
 ---
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -10,4 +10,15 @@ config.INTEGRATION_ID = config.INTEGRATION_ID.trim();
 config.CLUSTER_NAME = config.CLUSTER_NAME || 'Default cluster';
 config.IMAGE_STORAGE_ROOT = '/var/tmp';
 
+/**
+ * Important: we delete the following env vars because we don't want to proxy requests to the Kubernetes API server.
+ * The Kubernetes client library would honor the NO/HTTP/HTTPS_PROXY env vars.
+ */
+config.HTTPS_PROXY = process.env['HTTPS_PROXY'];
+config.HTTP_PROXY = process.env['HTTP_PROXY'];
+config.NO_PROXY = process.env['NO_PROXY'];
+delete process.env['HTTPS_PROXY'];
+delete process.env['HTTP_PROXY'];
+delete process.env['NO_PROXY'];
+
 export = config;
--- a/src/transmitter/index.ts
+++ b/src/transmitter/index.ts
@@ -1,9 +1,10 @@
 import * as needle from 'needle';
-import { NeedleResponse, NeedleHttpVerbs } from 'needle';
+import { NeedleResponse, NeedleHttpVerbs, NeedleOptions } from 'needle';
 import * as sleep from 'sleep-promise';
 import * as config from '../common/config';
 import logger = require('../common/logger');
 import { IDeleteWorkloadPayload, IDepGraphPayload, IWorkloadMetadataPayload, IResponseWithAttempts } from './types';
+import { getProxyAgent } from './proxy';
 
 const upstreamUrl = config.INTEGRATION_API || config.DEFAULT_KUBERNETES_UPSTREAM_URL;
 
@@ -72,10 +73,13 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
       'Client network socket disconnected before secure TLS connection was established',
     ],
   };
-  const options = {
+  const options: NeedleOptions = {
     json: true,
     compressed: true,
   };
+  if (config.HTTP_PROXY || config.HTTPS_PROXY) {
+    options.agent = getProxyAgent(config, url);
+  }
 
   let response: NeedleResponse | undefined;
   let attempt: number;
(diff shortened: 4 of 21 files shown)


src/common/config.ts
```diff
@@ -10,4 +10,15 @@ config.INTEGRATION_ID = config.INTEGRATION_ID.trim();
 config.CLUSTER_NAME = config.CLUSTER_NAME || 'Default cluster';
 config.IMAGE_STORAGE_ROOT = '/var/tmp';
 
+/**
+ * Important: we delete the following env vars because we don't want to proxy requests to the Kubernetes API server.
+ * The Kubernetes client library would honor the NO/HTTP/HTTPS_PROXY env vars.
+ */
+config.HTTPS_PROXY = process.env['HTTPS_PROXY'];
```
shaimendel, line 17: I wonder if we can do something a bit smarter - so customers insert values like `HTTPS_PROXY` but we pass them to the monitor as an env var called `SNYK_HTTPS_PROXY` (only in helm charts). so we will get it only as `config.HTTPS_PROXY` but it won't be part of the env vars, so no need to delete them

src/transmitter/index.ts
```diff
@@ -105,3 +107,38 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
...
   return {response, attempt};
 }
+
+/**
+ * Exported for testing
+ */
+export function getProxy(
+  config: Record<string, any>,
+  endpoint: string
+): string | undefined | never {
```
shaimendel, line 117: what is the `never` part?

src/common/config.ts
ivanstanev: Well that was my initial proposal 😛

src/transmitter/index.ts
ivanstanev: Being explicit that this function may throw, maybe that's something that only I would take note of 😆 
In languages like Java you got explicit exceptions in the method signature e.g. `throws IllegalAccessException`; this `never` here is a hint that the function throws

src/common/config.ts
shaimendel: I mean that the customers will still have the exact same interface of inserting `HTTPS_PROXY`, but we will map it to a `SNYK_HTTPS_PROXY` env var in the container

src/transmitter/index.ts
shaimendel: I'm not sure what's my opinion about that 😆 not sure that returning `never` will hint someone about an exception, for me it is a hint for an infinite loop

src/common/config.ts
ivanstanev: Just to make sure I understand, we will use `http_proxy`, `https_proxy`, etc. in the Helm chart, just in the `deployment.yaml` we will use `SNYK_HTTP_PROXY` for the name of the env var?

shaimendel: that's only my suggestion. what we actually want is customers to set `http_proxy` etc and for that to be "injected" to our config, right? 
not saying that's the best way to go, just a suggestion

src/transmitter/index.ts
ivanstanev: https://www.typescriptlang.org/docs/handbook/basic-types.html#never
It could be either 😆 To avoid ambiguity I can just remove that, doesn't seem helpful if it only raises questions

snyk-monitor-cluster-permissions.yaml
```diff
@@ -87,4 +87,7 @@ data:
   namespace: ""
   integrationApi: ""
   clusterName: ""
+  http_proxy: ""
```
Shesekino, line 90: any changes required for the OS operator?

package.json
```diff
@@ -9,6 +9,7 @@
     "test:system": "tap test/system --timeout=600",
     "test:integration:kind:yaml": "DEPLOYMENT_TYPE=YAML TEST_PLATFORM=kind CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
     "test:integration:kind:helm": "DEPLOYMENT_TYPE=Helm TEST_PLATFORM=kind CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
+    "test:integration:kind:helm:proxy": "DEPLOYMENT_TYPE=Proxy TEST_PLATFORM=kind CREATE_CLUSTER=true tap test/integration/kubernetes.test.ts --timeout=900",
```
Shesekino, line 12: this is getting out of hand 🙈 
if it's a deployment type then shouldn't it be ```test:integration:kind:proxy```? without the helm?

test/setup/deployers/index.ts
```diff
@@ -1,9 +1,11 @@
 import { yamlDeployer } from './yaml';
 import { operatorDeployer } from './operator';
 import { helmDeployer } from './helm';
+import { helmWithProxyDeployer } from './helm-with-proxy';
 
 export default {
   YAML: yamlDeployer,
   Operator: operatorDeployer,
   Helm: helmDeployer,
+  Proxy: helmWithProxyDeployer,
```
Shesekino, line 10: I love how your deployment types refactor is paying off! ❤️

snyk-monitor-cluster-permissions.yaml
ivanstanev: It's based on the Helm chart so it will automatically pick it up and we don't have to do anything else!

src/transmitter/index.ts
```diff
@@ -105,3 +109,58 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
...
   return {response, attempt};
 }
+
+/**
+ * Exported for testing
+ */
+export function getProxyAgent(
+  config: Record<string, any>,
+  endpoint: string
+): Agent | undefined {
```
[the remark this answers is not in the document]
ivanstanev: Good suggestion, let me know what you think of the adjusted changes!
I moved the `getProxyAgent()` to its own file, and the agent is now assigned only when HTTP/S_PROXY is set

src/transmitter/index.ts
```diff
@@ -105,3 +109,58 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
...
+      return httpsOverHttp({
+        proxy: {
+          host: httpsProxyUrl.hostname,
+          port: parseInt(httpsProxyUrl.port) || 80,
+        },
+      });
+
+    default:
+      const errorMessage = 'Unsupported protocol for proxying';
+      logger.error(errorMessage, { url, endpoint });
```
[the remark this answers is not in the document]
ivanstanev: Missed this one, but will fix it, and remove the last commit that checks the CI job.
What do you mean with the last comment, that the errorMessage shouldn't be assigned to a variable or something else?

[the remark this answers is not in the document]
ivanstanev: Fixed

src/transmitter/proxy.ts
```diff
@@ -0,0 +1,57 @@
...
+      return httpsOverHttp({
+        proxy: {
+          host: httpsProxyUrl.hostname,
+          port: parseInt(httpsProxyUrl.port) || 80,
+        },
+      });
+
+    default:
+      const errorMessage = 'Unsupported protocol for proxying';
+      logger.error({ url, endpoint }, errorMessage);
```
Shesekino, line 54:
```
logger.error({ url, endpoint }, 'Unsupported protocol for proxying');
throw new Error('Unsupported protocol for proxying');
```
wdyt?

Shesekino: we want to keep the plain text on the logger.error instruction so it's easier to reverse lookup from the logs

src/transmitter/proxy.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    // CIDR ranges are not supported, e.g. NO_PROXY=192.168.0.0/16.
+    // Wildcards are also not supported, e.g. NO_PROXY=*.mydomain.local
+    // HTTPS proxies (the connection to the proxy, not to the upstream) are not supported.
+  }
+
+  switch (url.protocol) {
+    case 'http:':
+      if (!config.HTTP_PROXY) {
```
shaimendel, line 29: what does this mean? to return the httpagent if no HTTP_PROXY mentioned

src/transmitter/proxy.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    case 'https:':
+      if (!config.HTTPS_PROXY) {
+        return HttpsAgent;
+      }
+      const httpsProxyUrl = new URL(config.HTTPS_PROXY);
+      return httpsOverHttp({
+        proxy: {
+          host: httpsProxyUrl.hostname,
+          port: parseInt(httpsProxyUrl.port) || 80,
```
shaimendel, line 48: is 80 a reasonable assumption here?
do we print the proxy details in case of errors in communication? or in general somewhere

src/transmitter/index.ts
```diff
@@ -72,10 +73,13 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
       'Client network socket disconnected before secure TLS connection was established',
     ],
   };
-  const options = {
+  const options: NeedleOptions = {
     json: true,
     compressed: true,
   };
+  if (config.HTTP_PROXY || config.HTTPS_PROXY) {
+    options.agent = getProxyAgent(config, url);
```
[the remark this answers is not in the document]
ivanstanev: I believe it's better to pass the dependencies of a function to it (DI-style), and it does help with testability this way 🙏

src/transmitter/proxy.ts
shaimendel: oh I got it :) never mind

src/transmitter/proxy.ts
ivanstanev: Great catch, should have been 443

src/transmitter/proxy.ts
ivanstanev: So Node.js has its own default HTTP/HTTPS agents which are used by networking libraries.
Here the function ensures to always return an agent. For http urls we return the default http agent; for https urls we return the default https agent.
Basically it means "use the default behaviour". The default agents are configured for normal, non-proxied requests

PR #446, opened 2020-04-20
Chore/rename depgraph type

ivanstanev:

### What this does

Rename a shortened type name to allow it to be searched more easily.

Add type hints to the mapping function constructing the dependency graph payload.
This helps us not introduce properties which are not part of the type. Previously we used "as IType" which could mask extra properties added to the object. Adding an explicit return type on the mapping function and removing the cast allows us to catch these errors at compile time.

### Screenshots

An example how this would now catch errors, previously we would have masked them with the cast:
Screenshot 2020-04-20 at 16 21 42

### DIFF
--- a/src/transmitter/types.ts
+++ b/src/transmitter/types.ts
@@ -33,7 +33,7 @@ export interface IKubernetesMonitorMetadata {
   namespace?: string;
 }
 
-export interface IDepGraphPayload {
+export interface IDependencyGraphPayload {
   imageLocator: IImageLocator;
   agentId: string;
   dependencyGraph?: string;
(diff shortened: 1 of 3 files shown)


src/transmitter/types.ts
```diff
@@ -33,7 +33,7 @@ export interface IKubernetesMonitorMetadata {
   namespace?: string;
 }
 
-export interface IDepGraphPayload {
+export interface IDependencyGraphPayload {
```
Shesekino, line 36: this is fine - just note that DepGraph is also used elsewhere
https://github.com/snyk/dep-graph

PR #459, opened 2020-04-28
Fix/when retry request

agranado2k:

### What this does

It has a small refactoring moving the condition that decide if k8s-monitor should or not retry a request. Also it includes a new error code that allows k8s-monitor decides if it should retry a request or not.

### Notes for the reviewer

This new error code was introduced because sometimes kubernetes cluster can have timeout caused by slow DNS lookup. And k8s-monitor should retry the request if it stills having remaining retries.

### More information

- [Jira ticket RUN-882](https://snyksec.atlassian.net/browse/RUN-882)

### DIFF
--- a/src/transmitter/index.ts
+++ b/src/transmitter/index.ts
@@ -122,5 +122,9 @@ function shouldRetryRequest(err: IRequestError, stillHaveRetries: boolean): bool
     return true;
   }
 
+  if (err.code === 'EAI_AGAIN') {
+    return true;
+  }
+
   return false;
 }
--- a/src/transmitter/index.ts
+++ b/src/transmitter/index.ts
@@ -8,6 +8,7 @@ import {
   IDependencyGraphPayload,
   IWorkloadMetadataPayload,
   IResponseWithAttempts,
+  IRequestError,
 } from './types';
 import { getProxyAgent } from './proxy';
 
@@ -73,10 +74,6 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
   const retry = {
     attempts: 3,
     intervalSeconds: 2,
-    networkErrorMessages: [
-      'socket hang up',
-      'Client network socket disconnected before secure TLS connection was established',
-    ],
   };
   const options: NeedleOptions = {
     json: true,
@@ -97,11 +94,7 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
         break;
       }
     } catch (err) {
-      if (!(
-        err.code === 'ECONNRESET' &&
-        retry.networkErrorMessages.includes(err.message) &&
-        stillHaveRetries
-      )) {
+      if (!shouldRetryRequest(err, stillHaveRetries)) {
         throw err;
       }
     }
@@ -113,4 +106,21 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
   }
 
   return {response, attempt};
+}
+
+function shouldRetryRequest(err: IRequestError, stillHaveRetries: boolean): boolean {
+  const networkErrorMessages: string[] = [
+    'socket hang up',
+    'Client network socket disconnected before secure TLS connection was established',
+  ];
+
+  if (!stillHaveRetries) {
+    return false;
+  }
+
+  if (err.code === 'ECONNRESET' && networkErrorMessages.includes(err.message)) {
+    return true;
+  }
+
+  return false;
 }
--- a/src/transmitter/types.ts
+++ b/src/transmitter/types.ts
@@ -72,3 +72,8 @@ export interface IResponseWithAttempts {
   response: NeedleResponse;
   attempt: number;
 }
+
+export interface IRequestError {
+  code: string;
+  message: string;
+}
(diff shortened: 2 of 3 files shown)


src/transmitter/index.ts
```diff
@@ -114,3 +112,15 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
 
   return {response, attempt};
 }
+
+function shouldRetryRequest(err: IRequestError, networkErrorMessages: string[], stillHaveRetries: boolean): boolean {
+  if (!stillHaveRetries) {
```
shaimendel, line 117: that's not a real review, just one comment about this function :) I don't like `else` when there is no need. just remove all the `else`:
```
if (something) {
  return something;
}

if (something else) {
  return something else;
}

return false;
```

agranado2k: Thanks @shaimendel! Definitely it's a good a good comment to get the code style!

src/transmitter/index.ts
```diff
@@ -114,3 +112,15 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
 
   return {response, attempt};
 }
+
+function shouldRetryRequest(err: IRequestError, networkErrorMessages: string[], stillHaveRetries: boolean): boolean {
```
ivanstanev, line 116: I think that `networkErrorMessages` is really only used within this `shouldRetryRequest()` function so it probably doesn't need to be passed as a variable any more. I think also currently the array gets allocated every time the function `retryRequest()` is called, so we probably don't need to do that! How about moving the `networkErrorMessages[]` as a const global variable in the current module?

src/transmitter/index.ts
```diff
@@ -114,3 +112,15 @@ async function retryRequest(verb: NeedleHttpVerbs, url: string, payload: object)
...
+  if (!stillHaveRetries) {
+    return false;
+  }
+
+  if (err.code === RequestCodeError.ECONNRESET && networkErrorMessages.includes(err.message)) {
```
ivanstanev, line 121: ➕ 1️⃣  on the Enum idea, I also think it's a really nice feature to use! However, in this current context I think it doesn't introduce benefits. It gives an extra layer of indirection (maybe you want to check what's the value of the enum?) and is a bit longer to type 😄 What do you think of just keeping it like before `(err.code === 'ECONNRESET')` as it is very explicit and easy to read?

test/system/kind.test.ts
```diff
@@ -135,7 +135,7 @@ tap.test('Kubernetes-Monitor with KinD', async (t) => {
   require('../../src');
 
   // TODO: replace with being event driven?
-  // will still need SOME timeout 
+  // will still need SOME timeout
```
ivanstanev, line 146: This change probably shouldn't be part of this commit. Maybe committed by accident since your linting tool has fixed it?
While it does remove an extra space, it's in a comment, so really not that relevant and it doesn't improve the quality of the codebase 😆

src/transmitter/types.ts
```diff
@@ -72,3 +72,12 @@ export interface IResponseWithAttempts {
   response: NeedleResponse;
   attempt: number;
 }
+
+export interface IRequestError {
+  code: string,
+  message: string
```
Shesekino, line 78: comma at the end of the line please

src/transmitter/types.ts
```diff
@@ -72,3 +72,12 @@ export interface IResponseWithAttempts {
...
+}
+
+export enum RequestCodeError {
+  ECONNRESET = 'ECONNRESET'
```
Shesekino: comma at the end of the line I think?

src/transmitter/index.ts
```diff
@@ -122,5 +122,9 @@ function shouldRetryRequest(err: IRequestError, networkErrorMessages: string[],
     return true;
   }
 
+  if (err.code === RequestCodeError.EAI_AGAIN) {
```
Shesekino, line 125: no need to specify "k8s-monitor" in the commit title; this is the Kubernetes-Monitor repo. unless specified otherwise, everything here is in the context of the kubernetes monitor.
so in this case, ```fix: retry transmitting data on EAI_AGAIN error codes``` might be more fitting

src/transmitter/types.ts
```diff
@@ -79,5 +79,6 @@ export interface IRequestError {
 }
 
 export enum RequestCodeError {
-  ECONNRESET = 'ECONNRESET'
+  ECONNRESET = 'ECONNRESET',
+  EAI_AGAIN = 'EAI_AGAIN'
```
Shesekino, line 83: comma at the end of the line

test/system/kind.test.ts
```diff
@@ -103,6 +103,14 @@ tap.test('Kubernetes-Monitor with KinD', async (t) => {
       message: 'socket hang up',
     });
 
+  nock('https://kubernetes-upstream.snyk.io')
```
Shesekino, line 106: nice!

src/transmitter/types.ts
agatakrajewska: I think it should be colons, like: 
```
export interface IRequestError {
  code: string;
  message: string;
}
```

agatakrajewska:
Yay, well done! 🙂 🥳 
Great approach with separate commits introducing refactoring first 👍
My main comment would also be about the commit messages, as @ivanstanev explained in detail, it's important to see why we are making certain changes, rather than pointing them out 🙂And also using just text without bullets and with punctuation +1️⃣from me as well, makes it super easy to read.

PR #469, opened 2020-05-04
chore: extract the condition for retry request to the new function

agranado2k:

### What this does

It has a small refactoring that moves the conditions that decides if k8s-monitor should or not retry a request using kubernetes API. It'll help include new logics that allow k8s-monitor retry for other errors.

### Notes for the reviewer

It's first step before include retry policies for errors in this ticket [RUN-818](https://snyksec.atlassian.net/browse/RUN-818).

### DIFF
--- a/src/supervisor/kuberenetes-api-wrappers.ts
+++ b/src/supervisor/kuberenetes-api-wrappers.ts
@@ -1,5 +1,6 @@
 import * as http from 'http';
 import * as sleep from 'sleep-promise';
+import { IRequestError } from './types';
 
 export const ATTEMPTS_MAX = 3;
 export const DEFAULT_SLEEP_SEC = 1;
@@ -13,20 +14,11 @@ export async function retryKubernetesApiRequest<ResponseType>(
     try {
       return await func();
     } catch (err) {
-      if (!(err.response)) {
+      if (!shouldRetryRequest(err, attempt)) {
         throw err;
       }
 
-      const response = err.response;
-      if (response.statusCode !== 429) {
-        throw err;
-      }
-
-      if (attempt === ATTEMPTS_MAX) {
-        throw err;
-      }
-
-      const sleepSeconds = calculateSleepSeconds(response);
+      const sleepSeconds = calculateSleepSeconds(err.response);
       await sleep(sleepSeconds * 1000);
     }
   }
@@ -47,4 +39,20 @@ export function calculateSleepSeconds(httpResponse: http.IncomingMessage): numbe
     }
   }
   return Math.min(sleepSeconds, MAX_SLEEP_SEC);
+}
+
+function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
+  if (attempt >= ATTEMPTS_MAX) {
+    return false;
+  }
+
+  if (!(err.response)) {
+    return false;
+  }
+
+  if (err.response.statusCode === 429) {
+    return true;
+  }
+
+  return false;
 }
--- a/src/supervisor/types.ts
+++ b/src/supervisor/types.ts
@@ -1,3 +1,4 @@
+import { IncomingMessage }  from 'http';
 import { AppsV1Api, BatchV1Api, BatchV1beta1Api, CoreV1Api, KubeConfig,
   V1ObjectMeta, V1OwnerReference, V1PodSpec } from '@kubernetes/client-node';
 
@@ -10,6 +11,10 @@ export enum WorkloadKind {
   CronJob = 'CronJob',
   ReplicationController = 'ReplicationController',
   Pod = 'Pod',
+}
+
+export interface IRequestError {
+  response?: IncomingMessage;
 }
 
 export interface IKubeObjectMetadata {


src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -1,32 +1,46 @@
...
 export async function retryKubernetesApiRequest<ResponseType>(
   func: IKubernetesApiFunction<ResponseType>
 ): Promise<ResponseType> {
   for (let attempt = 1; attempt <= ATTEMPTS_MAX; attempt++) {
     try {
       return await func();
     } catch (err) {
-      if (!(err.response)) {
-        throw err;
-      }
```
ivanstanev, line 18: This logic has now disappeared, which means that this new code on line 43:
```js
const sleepSeconds = calculateSleepSeconds(err.response);
```
Can now pass an undefined response to `calculateSleepSeconds` and cause an error!
So the logic of the function has changed, is it safe to do so?

If we keep this piece of code, then we don't have to return `undefined` in `getErrorCode()`. In fact, we also probably don't need a function there, since we can just directly reference `response.statusCode`!

src/supervisor/types.ts
```diff
@@ -1,3 +1,4 @@
+import * as http from 'http';
```
ivanstanev, line 1: What do you think of importing just the thing we need, e.g. `import { IncomingMessage } from 'http'`?

src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -1,32 +1,46 @@
...
+  return  err.response.statusCode;
+}
+
+function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
+  if (attempt === ATTEMPTS_MAX) {
+    return false;
+  }
+
+  const code = getErrorCode(err);
+  if (code === HTTP_STATUS_TOO_MANY_REQUESTS) {
```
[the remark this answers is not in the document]
jmickey: I actually prefer the `const` for this. Easier to understand what error we're looking for, and if I need to see the value I can just hover it in my editor. But I think this comes down to preference. I'll let @agranado2k decide what he prefers.

src/supervisor/kuberenetes-api-wrappers.ts
agranado2k: I was putting it in a way that it could help me to include new errors. But I think it would be good go a step back and rethink.

ivanstanev: I think that keeping that piece of code is more accurate, because now the code on line 37 (calculateSleepSeconds(...)) relies on some previous flow checking that the response is present, which is encapsulated in a function. If some day that check is removed, we'll probably break calculateSleepSeconds().

I think the existing flow is much more accurate:
- ensure the response is not falsy
- do something with the response
- do something else with the response

Otherwise, it becomes:
- do something with the response, but only if it's not empty
- do something else with the response (potentially also if it's not empty?)

I think we could catch these cases by changing the type to help us:
```js
export interface IRequestError {
  response?: IncomingMessage; // see how response is now optional
}
```

Maybe this would help surface the issue I'm trying to guard against?

In any case, it looks like we do handle an undefined response already: https://github.com/snyk/kubernetes-monitor/blob/89134966363adb3e160dcf84be206afc9223988b/src/supervisor/kuberenetes-api-wrappers.ts#L39
 so my suggestion would apply to any future places where we write similar code, to keep in mind not to rely too much on previous checks and handle the undefined check only in one place, at the top level 🙏

src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -1,32 +1,40 @@
...
 import * as sleep from 'sleep-promise';
+import { IRequestError } from './types';
 
 export const ATTEMPTS_MAX = 3;
 export const DEFAULT_SLEEP_SEC = 1;
 export const MAX_SLEEP_SEC = 5;
 type IKubernetesApiFunction<ResponseType> = () => Promise<ResponseType>;
 
+function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
+  if (attempt === ATTEMPTS_MAX) {
```
Shesekino, line 11: should probably consider the case where attempt > ATTEMPTS_MAX now being made possible by the extraction of this logic and "separating" it from the loop

src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -1,32 +1,40 @@
 import * as http from 'http';
 import * as sleep from 'sleep-promise';
+import { IRequestError } from './types';
 
 export const ATTEMPTS_MAX = 3;
 export const DEFAULT_SLEEP_SEC = 1;
 export const MAX_SLEEP_SEC = 5;
 type IKubernetesApiFunction<ResponseType> = () => Promise<ResponseType>;
 
+function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
```
Shesekino, line 10: usually nicer to put ```export```ed functions at the top to better convey the meaning of this module

src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -48,3 +40,19 @@ export function calculateSleepSeconds(httpResponse: http.IncomingMessage): numbe
   }
   return Math.min(sleepSeconds, MAX_SLEEP_SEC);
 }
+
+function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
+  if (attempt > ATTEMPTS_MAX) {
```
Shesekino, line 45: >= though?

agatakrajewska:
> Looks good! Left a bit of a lengthy comment 😅 but ✅ overall

It's def over 50 chars 🙂 Does it mean it will break somewhere??

ivanstanev:
> > Looks good! Left a bit of a lengthy comment 😅 but ✅ overall
> 
> It's def over 50 chars 🙂 Does it mean it will break somewhere??
> Also, great commit description!!

Haha, I meant this for a comment of mine on the code, it's resolved now

PR #472, opened 2020-05-06
Fix/include errors for k8s api calls

agranado2k:

### What this does

It includes new error cases to k8s API retry mechanism, and wrap two k8s API calls in the retry mechanism.

### Notes for the reviewer

It was broken in two commits. The first includes the new error cases to retry mechanism. The second wrap the k8s API calls in the retry mechanism.

### More information

- [Jira ticket RUN-818](https://snyksec.atlassian.net/browse/RUN-818)

### DIFF
--- a/src/supervisor/kuberenetes-api-wrappers.ts
+++ b/src/supervisor/kuberenetes-api-wrappers.ts
@@ -46,6 +46,14 @@ function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
     return false;
   }
 
+  if (err.code === 'ECONNREFUSED') {
+    return true;
+  }
+
+  if (err.code === 'ETIMEDOUT') {
+    return true;
+  }
+
   if (!(err.response)) {
     return false;
   }
--- a/test/system/kind.test.ts
+++ b/test/system/kind.test.ts
@@ -69,6 +69,30 @@ tap.test('Kubernetes-Monitor with KinD', async (t) => {
   ]);
 
   // Setup nocks
+  nock(/https\:\/\/127\.0\.0\.1\:\d+/, { allowUnmocked: true})
+    .get('/api/v1/namespaces')
+    .times(1)
+    .replyWithError({
+      code: 'ECONNREFUSED'
+    })
+    .get('/api/v1/namespaces')
+    .times(1)
+    .replyWithError({
+      code: 'ETIMEDOUT'
+    });
+
+  nock(/https\:\/\/127\.0\.0\.1\:\d+/, { allowUnmocked: true})
+    .get('/apis/apps/v1/namespaces/snyk-monitor/deployments')
+    .times(1)
+    .replyWithError({
+      code: 'ECONNREFUSED'
+    })
+    .get('/apis/apps/v1/namespaces/snyk-monitor/deployments')
+    .times(1)
+    .replyWithError({
+      code: 'ETIMEDOUT'
+    });
+
   nock('https://kubernetes-upstream.snyk.io')
     .post('/api/v1/workload')
     .times(1)
--- a/test/unit/supervisor/kubernetes-api-wrappers.test.ts
+++ b/test/unit/supervisor/kubernetes-api-wrappers.test.ts
@@ -46,6 +46,84 @@ tap.test('calculateSleepSeconds', async (t) => {
   );
 });
 
+tap.test('retryKubernetesApiRequest for ECONNREFUSED error', async (t) => {
+  const retryableErrorResponse = {code: 'ECONNREFUSED'};
+
+  await t.rejects(
+    () => kubernetesApiWrappers.retryKubernetesApiRequest(() => Promise.reject(retryableErrorResponse)),
+    'eventually throws on repeated retryable error responses',
+  );
+
+  let failures = 0;
+  const functionThatFailsJustEnoughTimes = () => {
+    if (failures < kubernetesApiWrappers.ATTEMPTS_MAX - 1) {
+      failures +=1;
+      return Promise.reject(retryableErrorResponse);
+    }
+    return Promise.resolve('egg');
+  };
+
+  const successfulResponse = await kubernetesApiWrappers.retryKubernetesApiRequest(functionThatFailsJustEnoughTimes);
+  t.equals(
+    successfulResponse,
+    'egg',
+    'keeps retrying on ECONNREFUSED as long as we don\'t cross max attempts',
+  );
+
+  failures = 0;
+  const functionThatFailsOneTooManyTimes = () => {
+    if (failures < kubernetesApiWrappers.ATTEMPTS_MAX) {
+      failures +=1;
+      return Promise.reject(retryableErrorResponse);
+    }
+    return Promise.resolve('egg');
+  };
+
+  await t.rejects(
+    () => kubernetesApiWrappers.retryKubernetesApiRequest(functionThatFailsOneTooManyTimes),
+    'failure more than the maximum, rejects, even for a retryable error response',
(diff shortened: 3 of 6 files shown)


test/unit/supervisor/kubernetes-api-wrappers.test.ts
```diff
@@ -46,6 +46,84 @@ tap.test('calculateSleepSeconds', async (t) => {
   );
 });
 
+tap.test('retryKubernetesApiRequest for ECONNREFUSED error', async (t) => {
+  const retryableErrorResponse = {code: 'ECONNREFUSED'};
+
+  t.rejects(
```
ivanstanev, line 52: I think we have to await here, otherwise `t.rejects()` creates a Promise that will run in the background, this can make the tests flaky

test/unit/supervisor/kubernetes-api-wrappers.test.ts
```diff
@@ -46,6 +46,84 @@ tap.test('calculateSleepSeconds', async (t) => {
...
+  failures = 0;
+  const functionThatFailsOneTooManyTimes = () => {
+    if (failures < kubernetesApiWrappers.ATTEMPTS_MAX) {
+      failures +=1;
+      return Promise.reject(retryableErrorResponse);
+    }
+    return Promise.resolve('egg');
+  };
+
+  t.rejects(
```
ivanstanev, line 82: Await here too! (Actually everywhere that we have .t.rejects() 😛 )

src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -46,11 +46,15 @@ function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
     return false;
   }
 
-  if (!(err.response)) {
-    return false;
```
ivanstanev, line 50: I think that the way it was before made a bit more sense because if we add another retry-on-status-code (e.g. on 503) then we have to again check if err.response is not falsy. 
```js
  if (err.response && err.response.statusCode === 429) {
    return true;
  }

  if (err.response.statusCode === 503) { // !!! err.response could be undefined
```

And then we'd have to do something like this which just repeats code...
```js
if (err.response && err.response.statusCode === 503) {
```

So something like before would guard us:
```js
  if (!(err.response)) {
    return false;
  }

  if (err.response.statusCode === 429) {
    // safe
  }

  if (err.response.statusCode === 503) {
    // safe
  }
```

test/system/kind.test.ts
```diff
@@ -69,6 +69,30 @@ tap.test('Kubernetes-Monitor with KinD', async (t) => {
   ]);
 
   // Setup nocks
+  nock(/https\:\/\/127\.0\.0\.1\:\d+/, { allowUnmocked: true})
```
ivanstanev, line 72: Just to note for our future tests we probably should think of a way to split this file into multiple tests as we are starting to pile a lot of logic here: retrying on errors to the API server, retrying on errors to kubernetes-upstream, checking that the payloads are as expected, etc. I think it's OK to do it like this now but we should keep in mind when we need to add more things here 🙏

src/supervisor/kuberenetes-api-wrappers.ts
agranado2k: I changed it because the next errors don't have `response`.  Like this one:
```ts
  if (err.code === 'ETIMEDOUT') {
    return true;
  }
```
However, I agree that won't become pretty. Maybe I can just invert? Include this case first and the case you mentioned after? Do you think it will be better?

test/system/kind.test.ts
agranado2k: Good point! Actually some point I thought to raise it in the channel, but after I waste so much time with nock I totally forget. But I totally agree! I definitely can do it now. I don't like.. next time I can do it! :)

src/supervisor/kuberenetes-api-wrappers.ts
ivanstanev: I think the way it was before was good if we had to add new status codes in the future. I think that won't be the case so let's leave it like it is now with your change 🙏 But if we did need to add a new code, we'll probably change the code the way it was before, so it becomes an unnecessary back-and-forth of the same change 😅

test/system/kind.test.ts
ivanstanev: Don't worry 😄 and it's not a waste of time, we learned something valuable and the time was spent towards improving the product and testing it properly

agranado2k: But one thing that increase the complexity to do it, it's because both tests will try to run in the same cluster. The options would be:
1) make the testes wait each other;
2) create different cluster for each test.

I prefer the second option. Any thoughts?

test/unit/supervisor/kubernetes-api-wrappers.test.ts
agranado2k: I can do it. Just confirm, because I copy this behaviour from this test. I think I should include `await` for all of them too, right?

```ts
tap.test('retryKubernetesApiRequest for retryable errors', async (t) => {
  const retryableErrorResponse = {response: {statusCode: 429}};

  t.rejects(
    () => kubernetesApiWrappers.retryKubernetesApiRequest(() => Promise.reject(retryableErrorResponse)),
    'eventually throws on repeated retryable error responses',
  );

  let failures = 0;
  const functionThatFailsJustEnoughTimes = () => {
    if (failures < kubernetesApiWrappers.ATTEMPTS_MAX - 1) {
      failures +=1;
      return Promise.reject(retryableErrorResponse);
    }
    return Promise.resolve('egg');
  };

  const successfulResponse = await kubernetesApiWrappers.retryKubernetesApiRequest(functionThatFailsJustEnoughTimes);
  t.equals(
    successfulResponse,
    'egg',
    'keeps retrying on 429 as long as we don\'t cross max attempts',
  );

  failures = 0;
  const functionThatFailsOneTooManyTimes = () => {
    if (failures < kubernetesApiWrappers.ATTEMPTS_MAX) {
      failures +=1;
      return Promise.reject(retryableErrorResponse);
    }
    return Promise.resolve('egg');
  };

  t.rejects(
    () => kubernetesApiWrappers.retryKubernetesApiRequest(functionThatFailsOneTooManyTimes),
    'failure more than the maximum, rejects, even for a retryable error response',
  );
});
```

test/system/kind.test.ts
ivanstanev: 👍 for option 2, we should always prefer running things in parallel I think

test/unit/supervisor/kubernetes-api-wrappers.test.ts
ivanstanev: Yes for sure, but my suggestion would be the improvement to come as a separate PR! 🙏 

Here's an example in another repo where we await: https://github.com/snyk/snyk-docker-plugin/blob/8f1d1e9e4a91fb642bf038d117ff548ea1a3b532/test/system/experimental.test.ts#L63-L71

src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -46,11 +46,19 @@ function shouldRetryRequest(err: IRequestError, attempt: number): boolean {
...
+  if (err.code === 'ETIMEDOUT') {
+    return true;
+  }
+
   if (!(err.response)) {
     return false;
   }
 
-  if (err.response.statusCode === 429) {
+  if (err.response && err.response.statusCode === 429) {
```
ivanstanev, line 61: Small nitpick: we already check for `err.response` in line 57

PR #475, opened 2020-05-07
fix: address TODO comments

agranado2k:

### What this does

 Address TODO comments, either removing it or improve code, once it was clear which action we should take about each one.

### Notes for the reviewer

Basically the TODOs are in two kinds of functions, the handlers and the readers.

For the handlers, we think it's ok we just remove the TODO and do nothing, we're already doing it in other handlers, and these handlers basically remove the workload from Homebase once it is removed from the cluster. If we have any problem, this workload would be removed from Homebase in 4 days anyway.

For the readers, it is worth including an info log with the workload name and namespace. Just because, in the future, if a customer has any problem, we will have some logs that will help us to investigate it.

### More information

- [Jira ticket RUN-545](https://snyksec.atlassian.net/browse/RUN-545)

### DIFF
--- a/src/supervisor/workload-reader.ts
+++ b/src/supervisor/workload-reader.ts
@@ -3,6 +3,7 @@ import { V1OwnerReference } from '@kubernetes/client-node';
 import * as kubernetesApiWrappers from './kuberenetes-api-wrappers';
 import { k8sApi } from './cluster';
 import { IKubeObjectMetadata, WorkloadKind } from './types';
+import logger = require('../common/logger');
 
 type IWorkloadReaderFunc = (
   workloadName: string,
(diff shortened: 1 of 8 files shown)


src/supervisor/workload-reader.ts
```diff
@@ -16,7 +17,8 @@ const deploymentReader: IWorkloadReaderFunc = async (workloadName, namespace) =>
 
   if (!deployment.metadata || !deployment.spec || !deployment.spec.template.metadata ||
       !deployment.spec.template.spec || !deployment.status) {
-    // TODO(ivanstanev): add logging to know when/if it happens!
+    logger.info({workloadName, namespace}, 'kubernetes api could not return workload');
```
ivanstanev, line 20: Just realised after the second review: this line seems to be repeating lots of times with identical structure, worth putting into a function? Something like `logIncompleteWorkload(workloadName, namespace);`, name up to you 🙏

PR #496, opened 2020-06-11
chore: skip Operator upgrade tests if no new version can be published

ivanstanev:

### What this does

Avoid testing Operator upgrades and pushing to a branch in https://github.com/snyk/community-operators if no new version was released.

### More information

- [Jira ticket RUN-989](https://snyksec.atlassian.net/browse/RUN-989)

### DIFF
--- a/.circleci/config/jobs/push_operator_to_community_operators.yml
+++ b/.circleci/config/jobs/push_operator_to_community_operators.yml
@@ -36,6 +36,14 @@ steps:
         echo "export NEW_OPERATOR_VERSION=${NEW_OPERATOR_VERSION}" >> $BASH_ENV
 
   - run:
+      name: End tests early if no new Operator is to be released
+      command: |
+        if [[ "${NEW_OPERATOR_VERSION}" == "${LAST_OPERATOR_VERSION}" ]]; then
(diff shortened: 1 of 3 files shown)


.circleci/config/jobs/push_operator_to_community_operators.yml
```diff
@@ -35,6 +35,14 @@ steps:
         NEW_OPERATOR_VERSION=${LATEST_TAG}
         echo "export NEW_OPERATOR_VERSION=${NEW_OPERATOR_VERSION}" >> $BASH_ENV
 
+  - run:
```
agatakrajewska, line 38: You don't want to have a var for this command as we use it twice in the same way?

ivanstanev: That's a good point but it kind of depends on the variables in the previous steps.
I guess we could have 2 commands:
- export last released operator and last released snyk-monitor
- this step which ends tests early if nothing has changed

Not sure if it's not gonna become harder to track... instead of having everything in one file you now have to look up commands somewhere else! This is also very specific only to the Operators so I'm not sure it's making things more maintainable.

PR #499, opened 2020-06-16
Test/add local registry coverage

agatakrajewska:

### What this does

We're adding missing coverage for local insecure registries support. Providing `registries.conf` config map `kubernetes-monitor` call `skopeo copy` images from insecure registries. 

### More information

- [RUN-980](https://snyksec.atlassian.net/browse/RUN-980)

### DIFF
--- a/test/fixtures/insecure-registries/python-local-deployment.yaml
+++ b/test/fixtures/insecure-registries/python-local-deployment.yaml
@@ -1,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: python-local
+  namespace: services
+  labels:
+    app.kubernetes.io/name: python-local
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: python-local
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: python-local
+    spec:
+      containers:
+      - image: kind-registry:5000/python:rc-buster
+        imagePullPolicy: Always
+        name: python-local
+        command: ['/bin/sleep']
+        args: ['9999999']
+      securityContext: {}
--- a/test/helpers/kubectl.ts
+++ b/test/helpers/kubectl.ts
@@ -64,6 +64,16 @@ export async function createSecret(
   console.log(`Created secret ${secretName}`);
 }
 
+export async function createConfigMap(
+  configMapName: string,
+  namespace: string,
+  filePath: string
+): Promise<void> {
+  console.log(`Creating config map ${configMapName} in namespace ${namespace}...`);
+  await exec(`./kubectl create configmap ${configMapName} -n ${namespace} --from-file=${filePath}`);
+  console.log(`Created config map ${configMapName}`);
+}
+
 export async function applyK8sYaml(pathToYamlDeployment: string): Promise<void> {
   console.log(`Applying ${pathToYamlDeployment}...`);
   await exec(`./kubectl apply -f ${pathToYamlDeployment}`);
@@ -143,6 +153,22 @@ export async function waitForServiceAccount(name: string, namespace: string): Pr
       await sleep(500);
     }
   }
+}
+
+export async function waitForJob(name: string, namespace: string): Promise<void> {
+  console.log(`Trying to find job ${name} in namespace ${namespace}`);
+  for (let attempt = 0; attempt < 60; attempt++) {
+    try {
+      await exec(`./kubectl get jobs/${name} -n ${namespace}`);
+    } catch (error) {
+      await sleep(1000);
+    }
+  }
+  console.log(`Found job ${name} in namespace ${namespace}`);
+
+  console.log(`Begin waiting for job ${name} in namespace ${namespace} to complete`);
+  await exec(`./kubectl wait --for=condition=complete jobs/${name} -n ${namespace} --timeout=240s`);
+  console.log(`Job ${name} in namespace ${namespace} is complete`);
 }
 
 async function getLatestStableK8sRelease(): Promise<string> {
--- a/test/integration/kubernetes.test.ts
+++ b/test/integration/kubernetes.test.ts
@@ -1,4 +1,5 @@
 import { CoreV1Api, KubeConfig, AppsV1Api } from '@kubernetes/client-node';
+import { exec } from 'child-process-promise';
 import setup = require('../setup');
 import * as tap from 'tap';
 import { WorkloadKind } from '../../src/supervisor/types';
@@ -21,6 +22,12 @@ tap.tearDown(async() => {
   console.log('Begin removing the snyk-monitor...');
   await setup.removeMonitor();
   console.log('Removed the snyk-monitor!');
+  console.log('Begin removing local container registry...');
+  await setup.removeLocalContainerRegistry();
+  console.log('Removed local container registry');
+  console.log('Begin removing "kind" network...');
+  await setup.removeUnusedKindNetwork();
+  console.log('Removed "kind" network');
 });
 
 // Make sure this runs first -- deploying the monitor for the next tests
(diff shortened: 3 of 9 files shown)


test/setup/platforms/kind.ts
```diff
@@ -59,7 +59,7 @@ async function download(osDistro: string): Promise<void> {
   } catch (error) {
     console.log('Downloading KinD...');
 
-    const url = `https://github.com/kubernetes-sigs/kind/releases/download/v0.7.0/kind-${osDistro}-amd64`;
+    const url = `https://github.com/kubernetes-sigs/kind/releases/download/v0.8.1/kind-${osDistro}-amd64`;
```
ivanstanev, line 64: Let's make this commit a chore as it affects only our tests, not worthy of a new feature to customers

test/fixtures/insecure-registries/python-local-deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+      app.kubernetes.io/name: python-local
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: python-local
+    spec:
+      containers:
+      - image: kind-registry:5000/python:rc-buster
+        imagePullPolicy: Always
+        name: nginx-local
```
ivanstanev, line 21: This container name is a bit misleading but not a big deal 😄

test/helpers/kubectl.ts
```diff
@@ -64,6 +64,16 @@ export async function createSecret(
   console.log(`Created secret ${secretName}`);
 }
 
+export async function createConfigmap(
```
ivanstanev, line 67: This could be capitalised `createConfigMap`

test/integration/kubernetes.test.ts
```diff
@@ -65,6 +69,17 @@ tap.test('snyk-monitor container started', async (t) => {
...
 
+tap.test('create local container registry', async (t) => {
+  console.log('Creating local container registry...');
+  await exec('docker run -d --restart=always -p "5000:5000" --name "kind-registry" registry:2');
+  await exec('docker network connect "kind" "kind-registry"');
+  
+  console.log('Pushing python:rc-buster image to the local registry');
+  await kubectl.applyK8sYaml('./test/fixtures/insecure-registries/push-to-registry-job.yaml');
+
+  t.pass('successfully pushed image to a local registry');
```
ivanstanev, line 80: WDYT of this test being "create local container registry and push an image" and we could also await the K8s Job here, this way the test is self-contained. Otherwise the test later at line 244 does a wait on the Job but folks may wonder where this job is coming from?

test/integration/kubernetes.test.ts
```diff
@@ -226,6 +241,45 @@ tap.test('snyk-monitor pulls images from a private ECR and sends data to kuberne
...
+    `api/v1/dependency-graphs/${integrationId}/${clusterName}/${namespace}/${deploymentType}/${deploymentName}`);
+  
+  t.ok('dependencyGraphResults' in depGraphResult,
+  'expected dependencyGraphResults field to exist in /dependency-graphs response');
+
+  /* Because of a bug in removeTagFromImage() func in src/scanner/images/index.ts, 
+  which chops off everything after ':' from the image name, we store a wrong image name 
+  and the result does not exist in the object referred below */
+  t.same(depGraphResult.dependencyGraphResults[imageName], null,
+  'expected result for image kind-registry:5000/python:rc-buster does not exist');
```
ivanstanev, line 280: Nice! Could we also show how it chops it off? So we prove the bug exists and when we fix it, we can remove this case? Otherwise the comparison to null could just mean that we didn't send the data!
```js
t.ok('kind-registry' in depGraphResult.dependencyGraphResults, 'BUG: the full image name is not stored in kubernetes-upstream');
```

test/integration/kubernetes.test.ts
```diff
@@ -314,3 +368,19 @@ tap.test('notify upstream of deleted pods that have no OwnerReference', async (t
...
+  const response = await k8sApi.readNamespacedDeployment(
+    'snyk-monitor',
+    'snyk-monitor',
+  );
+  const deployment = response.body;
+
+  validateSecureConfiguration(t, deployment);
+  validateVolumeMounts(t, deployment);
+  validateEnvironmentVariables(t, deployment);
+});
```
ivanstanev, line 386: Is this accidentally copied from somewhere?

test/setup/index.ts
```diff
@@ -6,6 +6,7 @@ import platforms, { getKubernetesVersionForPlatform } from './platforms';
 import deployers from './deployers';
 import { IImageOptions } from './deployers/types';
 import * as kubectl from '../helpers/kubectl';
+import { exec } from 'child-process-promise';
```
ivanstanev, line 9: Could this be at the top of the file?

test/setup/index.ts
```diff
@@ -37,6 +38,15 @@ export async function removeMonitor(): Promise<void> {
   }
 }
 
+export async function removeLocalContainerRegistry(): Promise<void> {
+  try {
+    await exec('docker rm kind-registry --force');
+    await exec('docker network prune --force');
```
ivanstanev, line 44: I think we should remove the `kind` network specifically and not _all_ unused networks on the system, just to be a bit more accurate and not touch other things that we didn't create

test/setup/index.ts
```diff
@@ -37,6 +38,15 @@ export async function removeMonitor(): Promise<void> {
   }
 }
 
+export async function removeLocalContainerRegistry(): Promise<void> {
```
ivanstanev, line 41: This also removes the docker network so the name is not entirely correct, I suggest we split it into two functions, one for each action. Note that currently if `docker rm kind-registry` fails then `docker network prune` won't be called

test/fixtures/insecure-registries/push-to-registry-job.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: push-to-registry
```
ivanstanev, line 4: WDYT of `copy-dockerhub-image-to-local-registry`? I know it's long but something like this can capture the essence of what this Job is doing without having to look at the details below

test/fixtures/insecure-registries/python-local-deployment.yaml
agatakrajewska: oops

test/integration/kubernetes.test.ts
agatakrajewska: So I had it like that (contained in one test) but the job takes some sweet time, so moving the 'wait' to the actual test reduces how long the tests take, as by the time we get to the test that job might have even completed. I thought that was more efficient, what do you think?

test/setup/index.ts
agatakrajewska: Sounds good

test/integration/kubernetes.test.ts
agatakrajewska: 😱

test/integration/kubernetes.test.ts
ivanstanev: Hmm I see! We could put a big warning here that the job should be awaited as it takes some time, and rename this message as it is a bit misleading because the image is not yet pushed to the registry.

There is a way to speed up things: we could build our own Docker image with skopeo installed on it, let's say it's named `snyk/runtime-fixtures:skopeo-copy` (similarly to `snyk/runtime-fixtures:tinyproxy) and then use this image in the Job with the simple command of `skopeo copy from to`, this way the Job should be fairly quick! Just a thought really 😄 

My dislike for this is the disassociation between where we create the job and where we await them - 100s of lines later. For comparison the other tests are somewhat in order - first test creates 5-6 deployments, next test makes sure the deployments are sent to the upstream.

agatakrajewska: Yeah, I totally agree, I really dislike the fact that setup is in 'test', because it's not a test and then that we are waiting for the job later on. 
I'll put a warning comment against it and I'm going to follow up with another PR with this optimisation to build our own image as it sounds good 👍 I'll add a note to the other task

ivanstanev: Thank you!

agatakrajewska: Np! :-)

test/setup/index.ts
agatakrajewska: Totally makes sense!

test/integration/kubernetes.test.ts
ivanstanev: I see this is resolved but the code is still here, why do we check the secure configuration? Isn't this tested similarly in a different place?

test/setup/index.ts
```diff
@@ -56,6 +73,7 @@ async function predeploy(integrationId: string): Promise<void> {
       'dockercfg.json': gcrDockercfg,
       integrationId,
     });
+    await createRegistriesConfigmap();
```
ivanstanev, line 76: Tiny nitpick: could be uppercase `Map`

test/integration/kubernetes.test.ts
agatakrajewska: no, it shouldn't be here at all, sorry 🤦‍♀️

PR #506, opened 2020-07-02
    fix: Pinning KinD version to 0.7.0 for OS3 tests

agatakrajewska:

### What this does

Pinning KinD binary version to 0.7.0, compatible with Kubernetes v1.11. We're using that specific version to mimic an OS3 cluster in our integration tests.

### DIFF
--- a/test/setup/index.ts
+++ b/test/setup/index.ts
@@ -133,7 +133,7 @@ export async function deployMonitor(): Promise<string> {
     await predeploy(integrationId);
 
     // TODO: hack, rewrite this
-    const imagePullPolicy = testPlatform === 'kind' ? 'Never' : 'Always';
+    const imagePullPolicy = testPlatform === 'kind' || testPlatform === 'openshift3' ? 'Never' : 'Always';
     const deploymentImageOptions: IImageOptions = {
       nameAndTag: remoteImageName,
       pullPolicy: imagePullPolicy,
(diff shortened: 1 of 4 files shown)


test/setup/platforms/kind.ts
```diff
@@ -7,7 +7,7 @@ const clusterName = 'kind';
 
 export async function setupTester(): Promise<void> {
   const osDistro = platform();
-  await download(osDistro);
+  await download(osDistro, 'v0.8.1');
```
shaimendel, line 10: why did you remove the version here? this parameter looks like mandatory in `download`

test/setup/index.ts
```diff
@@ -133,7 +133,7 @@ export async function deployMonitor(): Promise<string> {
     await predeploy(integrationId);
 
     // TODO: hack, rewrite this
-    const imagePullPolicy = testPlatform === 'kind' ? 'Never' : 'Always';
+    const imagePullPolicy = testPlatform === 'kind' || testPlatform === 'openshift3' ? 'Never' : 'Always';
```
agranado2k, line 136: A small comment. Just to avoid the code gets complicated. What do you think if we create a function, per platform, that specify which policy we should use to image pull?
Not sure if it will help us, or make the code harder to understand.

test/setup/platforms/kind.ts
agatakrajewska: No no, added a version here!

test/setup/index.ts
agatakrajewska: I think at this point I'd leave it here as I it would require extending the whole interface so we can do that once it does get complicated, wdyt?

PR #508, opened 2020-07-09
Fix/image name chopped for local registries

agatakrajewska:

### What this does

When image name was coming from a local registry eg. kind-registry:5000/python:rc-buster, everything after the first `:` was getting removed, this PR fixes that issue.

### More information

- [Jira ticket RUN-1019](https://snyksec.atlassian.net/browse/RUN-1019)

### Screenshots

image

### DIFF
--- a/src/scanner/images/index.ts
+++ b/src/scanner/images/index.ts
@@ -41,18 +41,31 @@ export async function removePulledImages(images: IPullableImage[]): Promise<void
 }
 
 // Exported for testing
-export function removeTagFromImage(imageWithTag: string): string {
-  return imageWithTag.split('@')[0].split(':')[0];
-}
-
-// Exported for testing
-export function getImageTag(imageWithTag: string): string {
-  const imageParts: string[] = imageWithTag.split(':');
-  if (imageParts.length === 2) { // image@sha256:hash or image:tag
-    return imageParts[1];
+export function getImageParts(imageWithTag: string) : {imageName: string, imageTag: string} {
+  // we're matching pattern: <registry:port_number>(optional)/<image_name>(mandatory)@<tag_identifier>(optional):<image_tag>(optional)
+  const regex = /((?:.*(:\d{4})?\/)?(?:[a-z0-9]+))([@|:].+)?/ig;
+  const groups  = regex.exec(imageWithTag);
+  
+  if(!groups){
+    logger.error({image: imageWithTag}, 'Image with tag is malformed, cannot extract valid parts');
+    return {imageName: imageWithTag, imageTag: ''};
   }
 
-  return '';
+  return {
+    imageName: groups[1],
+    imageTag: validateAndExtractImageTag(groups[3])
+  };
+}
+
+function validateAndExtractImageTag(imageTagGroup: string | undefined): string {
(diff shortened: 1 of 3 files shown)


src/scanner/images/index.ts
```diff
@@ -42,17 +42,33 @@ export async function removePulledImages(images: IPullableImage[]): Promise<void
...
+  return imageParts.imageTag;
+}
+
+function getImageParts(imageWithTag: string) : {imageName: string, imageTag: string} {
+  // we're matching pattern: <registry:port_number>(optional)/<image_name>(mandatory)@<tag_identifier>(optional):<image_tag>(optional)
+  const regex = /((?:.*(:\d{4})?\/)?(?:[a-z0-9]+))([@|:].+)?/ig;
+
+  const groups  = regex.exec(imageWithTag);
+  
+  if(!groups){
```
shaimendel, line 61: do we actually want to "lose" this image scan in this case? wdyt for example to fallback to the old flow in this case instead?

src/scanner/images/index.ts
```diff
@@ -42,17 +42,33 @@ export async function removePulledImages(images: IPullableImage[]): Promise<void
...
   }
 
-  return '';
+  const imageTagParts: string[] | undefined = groups[3] ? groups[3].split(':') : undefined;
+
+  return {
+    imageName: groups[1],
+    //checking if tag malformed, valid formats: image@sha256:hash or image:tag
+    imageTag: imageTagParts && imageTagParts.length === 2 ? imageTagParts[1] : ''
+  };
```
shaimendel, line 71: I find this flow a bit hard to read and understand, might worth a small refactor

agatakrajewska: fair enough! 👍

src/scanner/images/index.ts
agatakrajewska: You are right, what I really should be doing here is logging an error (I think), but not sure what we want to return in this case, old flow would just return the image name as it comes, as before we had `imageWithTag.split('@')[0].split(':')[0]`, so if the image name is completely malformed, say doesn't contain `@` or `:`, we would just get the `imageWithTag` sting. Wdyt of the change?

src/scanner/images/index.ts
agatakrajewska: I have refactored it, so hopefully now it's a lot clearer

src/scanner/images/index.ts
shaimendel: looks good :) what will happen in this case in registry etc? just to make sure nothing crashes 😆

agatakrajewska: So I think we should be getting exactly the same result as before, we've got quite a few tests covering that workflow, but only deploying images with valid names. I'm not sure if this situation would even take place as if the image name/ tag is malformed, would it just not be available as a deployment we can scan because it couldn't pull a valid image?!

agatakrajewska: In the integration tests we're covering few image names possibilities: Image with a tag, image from a local registry, private registry, image with a sha, I would say we are covered but if I could test it somehow to make sure, I'm open for suggestions 🤔

PR #533, opened 2020-08-11
chore: push snyk operator to operatorhub.io

agranado2k:

### What this does

To be able to publish the Snyk Operator on OperatorHub.io we need to copy the operator folder from kubernetes monitor repo to upstream-community-operator folder into community-operators repo.
Because we're already doing it for the community-operator folder, we're creating two different branches to make it easier and avoid any conflicts.

### Notes for the reviewer

It's pushing the branch to [snyk/community-operators](https://github.com/snyk/community-operators/branches) repo. And it opens the PR for both `commuinty-operators` (openshift) and `upstream-community-opearators` (kubernetes).

### More information

- [RUN-1017](https://snyksec.atlassian.net/browse/RUN-1017)

### Screenshots
The branches on snyk/community-operator repo.
Screenshot 2020-08-11 at 15 51 39

Screenshot 2020-08-11 at 18 11 03

Screenshot 2020-08-11 at 18 12 04

Screenshot 2020-08-11 at 18 13 53


test/setup/platforms/kind-olm.ts
```diff
@@ -0,0 +1,31 @@
...
+  // await kubectl.waitForDeployment('operatorhubio-catalog', 'olm');
+  await kubectl.waitForDeployment('packageserver', 'olm');
+
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/01_namespace.yaml');
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/03_operatorsource.crd.yaml');
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/04_service_account.yaml');
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/05_role.yaml');
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/06_role_binding.yaml');
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/07_upstream_operatorsource.cr.yaml');
+  await kubectl.applyK8sYaml('./test/fixtures/operator/marketplace/08_operator.yaml');
```
ivanstanev, line 29: We can probably apply these from GitHub too, if we pick them from a specific commit (to ensure whatever is on develop/main doesn't change) 🤔

test/helpers/kubectl.ts
```diff
@@ -155,6 +169,18 @@ export async function waitForServiceAccount(name: string, namespace: string): Pr
   }
 }
 
+export async function waitForCRD(name: string): Promise<void> {
+  // TODO: add some timeout
+  while (true) {
```
ivanstanev, line 174: Can we add it now? Otherwise I feel we will never come back to this 😄 I think we should avoid adding more TODOs unless we really cannot fix it now

test/setup/index.ts
```diff
@@ -62,18 +66,23 @@ async function createEnvironment(): Promise<void> {
...
   try {
-    const namespace = 'snyk-monitor';
-    await kubectl.createNamespace(namespace);
-
     const secretName = 'snyk-monitor';
+    console.log(`Creating namespace ${namespace} and secret ${secretName}`);
+
+    try {
+      await kubectl.createNamespace(namespace);
+    } catch (error) {
```
ivanstanev, line 76: Why would this fail, how can it exist already?

test/setup/deployers/operator.ts
```diff
@@ -1,19 +1,53 @@
+import * as sleep from 'sleep-promise';
 import { IDeployer, IImageOptions } from './types';
 import * as kubectl from '../../helpers/kubectl';
 
+const PULL_KUBERNETES_OPERATOR_EVENT = 'docker.io/snyk/kubernetes-operator';
```
ivanstanev, line 5: Is there a little bit more info we can add to this event, not just the image name? Or you think this gives us enough confidence that we have the right event?

test/setup/deployers/operator.ts
```diff
@@ -1,19 +1,53 @@
...
 export const operatorDeployer: IDeployer = {
   deploy: deployKubernetesMonitor,
 };
 
+async function seekEvent(event: string): Promise<boolean> {
+  const events = await kubectl.getEvents('marketplace');
```
ivanstanev, line 11: Function name looks generic but we use a hardcoded OLM namespace, could we explicitly also pass the namespace to `seekEvent()`?

test/setup/deployers/operator.ts
```diff
@@ -1,19 +1,53 @@
...
+  const foundEvent = events.indexOf(event) > -1;
+
+  return foundEvent;
+}
+
+async function waitToDeployKubernetesOperator(): Promise<void> {
+  while(true) {
```
ivanstanev, line 18: Could we also give it a max limit here?

test/setup/deployers/operator.ts
```diff
@@ -1,19 +1,53 @@
...
-  // the Operator can start processing the custom resource.
-  await kubectl.waitForDeployment('snyk-operator', 'snyk-monitor');
+    // Await for the Operator to become available, only then
+    // the Operator can start processing the custom resource.
+    await kubectl.waitForDeployment('snyk-operator', 'marketplace');
+    await kubectl.waitForCRD('snykmonitors.charts.helm.k8s.io');
+    await kubectl.applyK8sYaml('./test/fixtures/operator/custom-resource-k8s.yaml');
+    await waitToDeployKubernetesOperator();
+    await kubectl.waitForDeployment('snyk-operator', 'marketplace');
+  }
```
ivanstanev, line 51: Hmm it almost feels like we could have two distinct deployers - `openshift-deployer.ts` and `olm-deployer.ts`, they seem distinct enough to warrant their own deployment method! This would avoid doing these `if` checks and doing completely different things based on the namespaces we have in the cluster.

test/setup/platforms/index.ts
```diff
@@ -16,6 +17,8 @@ interface IPlatformSetup {
   clean: () => Promise<void>;
   // set up host requirements specific to this platform
   setupTester: () => Promise<void>;
+  // namespace where snyk-monitor will be deployed
+  namespace: () => Promise<string>;
```
ivanstanev, line 21: I think this isn't the right place to put it - the file relates only to the _platform setup_ and here we are mixing deployment with it ("namespace where snyk-monitor will be deployed")!

test/setup/index.ts
agranado2k: For the case when we try to create the `marketplace` namespace, it already exists and it will raise an error. To avoid that I put it around the try-catch statement.

test/setup/deployers/operator.ts
agranado2k: What do you think if I include a comment showing an example of the event we're looking for? Because it's exactly the relevant information that we need to look for in the event to make sure that it has loaded the correct operator image.

test/setup/platforms/index.ts
agranado2k: Hum... I put it there because the namespace will be chosen depending on the platform that we're using. Maybe the comment is really bad. It might be `The namespace where snyk-monitor will be deployed depends on the platform is running the test`.

test/setup/deployers/operator.ts
ivanstanev: That works too 😄 Just not sure why not put the whole event object in the string, rather than a portion of the event and the whole event in a comment

test/setup/platforms/index.ts
ivanstanev: I just think the whole concept of this module is to deal with the platform - e.g. raising a cluster, destroying it, etc. It shouldn't have the notion of namespaces because that's a deployment concern. And for deploying we have a separate `deployers` module.  I think some of our abstractions are starting to mix together that's all! If there isn't an obvious other place to put this function (figuring out which namespace we should deploy) then it could stay but to me it feels wrong 😬

.circleci/config/@config.yml
```diff
@@ -38,6 +38,14 @@ workflows:
...
             - build_image
           <<: *main_branches_filter
+      - integration_tests_proxy:
+          requires:
+            - build_image
+          <<: *main_branches_filter
+      - integration_tests_proxy:
+          requires:
+            - build_image
+          <<: *main_branches_filter
```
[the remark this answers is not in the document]
agranado2k: There is no reason. I think I took along it. I'm removing.

test/README.md
```diff
@@ -83,6 +83,14 @@ This test runs whenever we commit to our `staging` branch, and at the moment may
 
 Run with `npm run test:integration:openshift4`.
 
+### KinD with OLM ###
+
+This test helps us ensure our operator is installable and functioning on a plain k8s cluster.
+
+This test runs whenever we commit to any branch.
```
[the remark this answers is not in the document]
agranado2k: I agree! We should get some time to think how we could improve that. However now it's a limitation with `OperatorSource`. Maybe we can think about how can we create our own Catalog for test. But I think it'll take some time to do it! :)

test/setup/deployers/operator-openshift.ts
```diff
@@ -10,10 +10,8 @@ async function deployKubernetesMonitor(
 ): Promise<void> {
   await kubectl.applyK8sYaml('./test/fixtures/operator/operator-source.yaml');
   await kubectl.applyK8sYaml('./test/fixtures/operator/installation.yaml');
-
   // Await for the Operator to become available, only then
   // the Operator can start processing the custom resource.
-  await kubectl.waitForDeployment('snyk-operator', 'snyk-monitor');
-
+  await kubectl.waitForDeployment('snyk-operator', 'openshift-marketplace');
```
ivanstanev, line 15: I think this isn't quite right, we want the Operator to appear in the `snyk-monitor` namespace -- this is what would happen if the user installed it through the OpenShift UI

test/setup/platforms/kind-olm.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  // OLM installation: https://github.com/operator-framework/community-operators/blob/master/docs/testing-operators.md#2-install-olm
+
+  const operatorLifecycleManagerVersion = '0.15.1'; // https://github.com/operator-framework/operator-lifecycle-manager/releases/tag/0.15.1
+  await kubectl.applyK8sYaml(`https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${operatorLifecycleManagerVersion}/crds.yaml`);
+  await sleep(5000); // give enough time to k8s to apply the previous yaml
+  await kubectl.applyK8sYaml(`https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${operatorLifecycleManagerVersion}/olm.yaml`);
+  await kubectl.waitForDeployment('catalog-operator', 'olm');
+  await kubectl.waitForDeployment('olm-operator', 'olm');
+  // await kubectl.waitForDeployment('operatorhubio-catalog', 'olm');
```
ivanstanev, line 16: Do we need this catalog, or it is optional?

test/setup/deployers/operator-openshift.ts
agranado2k: Definitely! I'm making sure that all the integration tests keep running and passing.

test/setup/platforms/kind-olm.ts
agranado2k: We don't need it. I just forgot to delete it!

.circleci/config/jobs/integration_tests_operator_on_k8s.yml
```diff
@@ -0,0 +1,27 @@
...
+    name: Create temporary directory for logs
+- run:
+    command: |
+      export KUBERNETES_MONITOR_IMAGE_NAME_AND_TAG=$(./scripts/circleci-jobs/setup-integration-tests.py)
+      .circleci/do-exclusively --branch staging --job ${CIRCLE_JOB} npm run test:integration:kindolm:operator
+    name: Operator integration tests on plain k8s
+- run:
+    name: Delete Operator from Quay
+    command: |
+      ./scripts/operator/delete-operator-from-quay.sh
```
shaimendel, line 19: just making sure there is no race condition here with the regular operator tests (because they both delete the operator)

.circleci/config/jobs/push_operator_to_community_operators.yml
```diff
@@ -51,43 +51,23 @@ steps:
...
-        git commit -s -m "Upgrade snyk-operator to version ${NEW_OPERATOR_VERSION}"
-        git push --set-upstream origin --force snyk/snyk-operator-v${NEW_OPERATOR_VERSION}
-
+        COMMUNITY_FOLDER_LOCATION="community-operators"
+        ./scripts/operator/push-operator-to-snyk-upstream.sh "${CURRENT_DIRECTORY}" "${COMMUNITY_FOLDER_LOCATION}"
+  - run:
+      name: Open a Pull Request to the operator-framework/community-operators repository for community-operators
+      command: |
+        COMMUNITY_FOLDER_LOCATION="community-operators"
+        python ./scripts/operator/raise_pr_to_community_operators_from_our_fork.py "${COMMUNITY_FOLDER_LOCATION}" "${NEW_OPERATOR_VERSION}"
```
shaimendel, line 61: will we be notified on slack about that? ideally with the PR link

agranado2k: Yep! In the command below `./scripts/slack/notify_success_operator_push.py "${NEW_OPERATOR_VERSION}" "${COMMUNITY_FOLDER_LOCATION}"` it will send a Slack message to us.
```python
   data = {
      'attachments':
      [
        {
          'color': '#7CD197',
          'fallback': 'Build Notification: ' + circle_build_url,
          'title': 'Snyk Operator Pushed to GitHub repo snyk/community-operators',
          'text': 'Branch *snyk/snyk-operator-v' + operator_version + '* is ready for publishing to the ' + upstream_community + '.\n' +
            'https://github.com/operator-framework/community-operators/compare/master...snyk:snyk/' + upstream_community + '/snyk-operator-v' + operator_version
        }
      ]
    }
```

.circleci/config/jobs/integration_tests_operator_on_k8s.yml
agranado2k: Actually it's one point in the document that Ivan has created about tech debts on operators. For now we think it's ok, because it will only delete the operator in the test.  Because the integration tests take a long time to run and we need it only in the early part of the test to pull the image, it's ok.
However we think we can improve it in the future. Maybe create another pipeline step that will run after all the integration tests (doesn't matter if it has failed or not).

.circleci/config/jobs/push_operator_to_community_operators.yml
shaimendel: this is about us pushing the code, it doesn't open the PR itself I think (the text of this message at least says that). do we actually open the PR?

PR #547, opened 2020-08-20
[RUN-1105] chore: fail the build if deployer returns an error

shaimendel:

### What this does

fail the build if deployer returns an error

### More information

- [Jira ticket RUN-1105](https://snyksec.atlassian.net/browse/RUN-1105)

### DIFF
--- a/.circleci/config/jobs/@jobs.yml
+++ b/.circleci/config/jobs/@jobs.yml
@@ -325,7 +325,7 @@ deploy_dev:
           LATEST_TAG_WITH_V=`git describe --abbrev=0 --tags ${CIRCLE_SHA1}` &&
           LATEST_TAG=${LATEST_TAG_WITH_V:1}-approved &&
           ./scripts/slack/notify_deploy.py $LATEST_TAG dev &&
-          curl -i -H "Accept: application/json" -H "Content-Type: application/json" \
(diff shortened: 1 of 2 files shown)


.circleci/config/jobs/@jobs.yml
```diff
@@ -325,7 +325,7 @@ deploy_dev:
           LATEST_TAG_WITH_V=`git describe --abbrev=0 --tags ${CIRCLE_SHA1}` &&
           LATEST_TAG=${LATEST_TAG_WITH_V:1}-approved &&
           ./scripts/slack/notify_deploy.py $LATEST_TAG dev &&
-          curl -i -H "Accept: application/json" -H "Content-Type: application/json" \
+          curl -s --fail --show-error -i -H "Accept: application/json" -H "Content-Type: application/json" \
```
agatakrajewska, line 328: cool, so I see `show-error` will send the error to `stderr` and that's how it will fail in circle? :-)

agatakrajewska (approved):

.circleci/config/jobs/@jobs.yml
shaimendel: no, the show-error is only to print the error to stdout :) the `--fail` is the one that makes curl return with a failure if the http status code is not 200

PR #572, opened 2020-10-15
feat: control log level verbosity

dkontorovskyy:
- [X] Documentation written [ℹ︎](https://github.com/snyk/general/wiki/Documentation)

### What this does
Control log level verbosity

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ no_proxy: *.example.local,*.other.global,192.168.0.0/16
 no_proxy: long.domain.name.local,example.local
 ```
 
+## Changing log level ##
+
+To lower `snyk-monitor`'s logging verbosity `log_level` value could be set to one of these options:
+* `'WARN'`
+* `'ERROR'`
+
+By default, `log_level` is `'INFO'`.
+
 ## Terms and conditions ##
 
 *The Snyk Container Kubernetes integration uses Red Hat UBI (Universal Base Image).*
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -58,6 +58,8 @@ spec:
             value: {{ .Values.https_proxy }}
           - name: NO_PROXY
             value: {{ .Values.no_proxy }}
+          - name: LOG_LEVEL
+            value: {{ .Values.log_level }}
           resources:
             requests:
               cpu: {{ .Values.requests.cpu }}
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -1,6 +1,13 @@
 import * as bunyan from 'bunyan';
 import config = require('./config');
 
+try {
+  // Validate if LOG_LEVEL has valid bunyan logging level
+  config.LOGGING.level = bunyan.resolveLevel(process.env.LOG_LEVEL || config.LOGGING.level);
+} catch(e) {
+  console.error(`Log level "${process.env.LOG_LEVEL}" is not valid logging level. Falling back to "INFO"`);
+}
+
 const logger: bunyan = bunyan.createLogger({
   name: config.LOGGING.name,
   level: config.LOGGING.level,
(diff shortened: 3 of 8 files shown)


snyk-monitor/templates/deployment.yaml
```diff
@@ -58,6 +58,8 @@ spec:
             value: {{ .Values.https_proxy }}
           - name: NO_PROXY
             value: {{ .Values.no_proxy }}
+          - name: LOG_LEVEL
+            value: {{ .Values.log_level }}
```
agranado2k, line 62: I guess you need to include the same options to the helm templates [here](https://github.com/snyk/kubernetes-monitor/blob/ecca7f5ce26fc8926660871ed07896a3109199be/snyk-monitor/templates/deployment.yaml#L59) for `snyk-monitor/templates/deployment.yaml`

agranado2k (commented):
I have a comment! :)
Also I believe we should have at least one test to make sure it's working in the way we expect!

snyk-monitor/templates/deployment.yaml
dkontorovskyy: Your link is pointing to the same place it's in PR, am I missing smth?

README.md
```diff
@@ -140,6 +140,14 @@ no_proxy: *.example.local,*.other.global,192.168.0.0/16
 no_proxy: long.domain.name.local,example.local
 ```
 
+## Changing default log level ##
+
+To lower `snyk-monitor`'s logging verbosity `log_level` value could be set to one of these options:
+* `'WARN'`
+* `'ERROR'`
+
+By default, `log_level` is `'INFO'`
```
shaimendel, line 149: I'm really sorry about this - but can you please add a `.` in the end? :) to match the style of the docs

README.md
```diff
@@ -140,6 +140,14 @@ no_proxy: *.example.local,*.other.global,192.168.0.0/16
 no_proxy: long.domain.name.local,example.local
 ```
 
+## Changing default log level ##
```
shaimendel, line 143: maybe remove the `default` part?

shaimendel, line 143: need to add this also to https://github.com/snyk/kubernetes-monitor/blob/staging/snyk-monitor/README.md with concrete helm example (this file is for YAML installation)

src/common/logger.ts
```diff
@@ -1,6 +1,17 @@
 import * as bunyan from 'bunyan';
 import config = require('./config');
 
+if (process.env.LOG_LEVEL) {
+  try {
+    // Validate if LOG_LEVEL has valid bunyan logging level
+    bunyan.resolveLevel(process.env.LOG_LEVEL);
+  } catch {
+    delete process.env.LOG_LEVEL;
```
shaimendel, line 9: I know it is a problematic context but can we log something here? so the customer knows what happened

src/common/logger.ts
```diff
@@ -1,6 +1,17 @@
...
+  } finally {
+    config.LOGGING.level = process.env.LOG_LEVEL || config.LOGGING.level;
```
shaimendel, line 11: can't this be moved to after the `bunyan.resolveLevel` call in the try clause? so we don't need finally at all.
```js
try {
    // Validate if LOG_LEVEL has valid bunyan logging level
    bunyan.resolveLevel(process.env.LOG_LEVEL);
    config.LOGGING.level = process.env.LOG_LEVEL;
  } catch {
    delete process.env.LOG_LEVEL;
  }
```

also - not sure why we need/want to delete the env var

src/common/logger.ts
```diff
@@ -1,6 +1,13 @@
 import * as bunyan from 'bunyan';
 import config = require('./config');
 
+try {
+  // Validate if LOG_LEVEL has valid bunyan logging level
+  config.LOGGING.level = bunyan.resolveLevel(process.env.LOG_LEVEL || config.LOGGING.level);
+} catch(e) {
+  console.error(`Log level "${process.env.LOG_LEVEL}" is not valid logging level. Falling back to "INFO"`);
```
shaimendel, line 8: That's good, just need to take into account that it won't match the "schema" of the rest of the logs, but I don't think we have a choice here

shaimendel (approved):

PR #580, opened 2020-10-20
feat: pull image by SHA instead of tag

dkontorovskyy:

### What this does
Pull image by SHA instead of the tag

### DIFF
--- a/src/scanner/images/index.ts
+++ b/src/scanner/images/index.ts
@@ -3,31 +3,31 @@ import * as plugin from 'snyk-docker-plugin';
 
 import logger = require('../../common/logger');
 import { pull as skopeoCopy, getDestinationForImage } from './skopeo';
-import { IPullableImage } from './types';
+import { IPullableImage, IScanImage } from './types';
 import { IStaticAnalysisOptions, StaticAnalysisImageType, IScanResult, IPluginOptions } from '../types';
 
 export async function pullImages(images: IPullableImage[]): Promise<IPullableImage[]> {
   const pulledImages: IPullableImage[] = [];
 
   for (const image of images) {
-    const {imageName, fileSystemPath} = image;
+    const {imageWithDigest, fileSystemPath} = image;
     if (!fileSystemPath) {
       continue;
     }
 
     try {
-      await skopeoCopy(imageName, fileSystemPath);
+      await skopeoCopy(imageWithDigest, fileSystemPath);
       pulledImages.push(image);
     } catch (error) {
-      logger.error({error, image: imageName}, 'failed to pull image');
+      logger.error({error, image: imageWithDigest}, 'failed to pull image');
     }
   }
 
   return pulledImages;
 }
 
-export function getImagesWithFileSystemPath(images: string[]): IPullableImage[] {
-  return images.map((image) => ({ imageName: image, fileSystemPath: getDestinationForImage(image) }));
+export function getImagesWithFileSystemPath(images: IScanImage[]): IPullableImage[] {
+  return images.map((image) => ({ ...image, fileSystemPath: getDestinationForImage(image.imageName) }));
 }
 
 export async function removePulledImages(images: IPullableImage[]): Promise<void> {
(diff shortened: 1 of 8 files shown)


src/scanner/images/index.ts
```diff
@@ -3,31 +3,31 @@ import * as plugin from 'snyk-docker-plugin';
...
-import { IPullableImage } from './types';
+import { IPullableImage, IScanImage } from './types';
 import { IStaticAnalysisOptions, StaticAnalysisImageType, IScanResult, IPluginOptions } from '../types';
 
 export async function pullImages(images: IPullableImage[]): Promise<IPullableImage[]> {
   const pulledImages: IPullableImage[] = [];
 
   for (const image of images) {
-    const {imageName, fileSystemPath} = image;
+    const {imageDigest, fileSystemPath} = image;
```
ivanstanev, line 13: Could we preserve the `imageName` so that at least we can log it in case of errors? Might be hard to understand which image is problematic just by looking at its digest 🙏

src/transmitter/types.ts
```diff
@@ -25,6 +25,7 @@ export interface IWorkloadMetadata {
 
 export interface IImageLocator extends IWorkloadLocator {
   imageId: string;
+  imageDigest: string;
```
ivanstanev, line 28: Is it possible to name it `imageWithDigest` just like it is on the `IScanImage`? I think just `imageDigest` conveys the message that it has the hash/digest but actually it also has the image name in here

I see that we have previously messed it up a little bit, adding fields like `imageName` which contain both the name _and the tag_...

ivanstanev (commented):
Looks amazing 😄 

I have a small suggestion about the commit message - could we also describe _why_ we are doing this change? Would be useful for when someone needs to go over the commit history to understand the reason for switching to this way of pulling

src/scanner/images/index.ts
```diff
@@ -41,15 +41,15 @@ export async function removePulledImages(images: IPullableImage[]): Promise<void
...
   // we're matching pattern: <registry:port_number>(optional)/<image_name>(mandatory):<image_tag>(optional)@<tag_identifier>(optional)
   // extracted from https://github.com/docker/distribution/blob/master/reference/regexp.go
   const regex = /^((?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?\/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:\/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][A-Fa-f0-9]{32,}))?$/ig;
   const groups  = regex.exec(imageWithTag);
-  
+
   if(!groups){
     logger.error({image: imageWithTag}, 'Image with tag is malformed, cannot extract valid parts');
-    return {imageName: imageWithTag, imageTag: ''};
+    return { imageName: imageWithTag, imageTag: '', imageDigest: '' };
```
ivanstanev, line 52: This seems like a bit of an error if we returned an empty `imageDigest` considering the rest of the code flow is going to depend on it! Is it possible to fall back to `imageName`/`imageWithTag` when `imageDigest` is empty? Otherwise we might show users empty results or something further up the stack is going to throw when it cannot parse the digest.

dkontorovskyy: This verifies this kind of cases, when name is malformed

https://github.com/snyk/kubernetes-monitor/blob/40c8ff4f1f2aabbb0eef861862ef59d6105b4cb4/test/unit/scanner/images.test.ts#L84

ivanstanev: Correct, but it also returns an empty `imageDigest` when it's malformed. So what will happen if we call `skopeoCopy(emptyImageDigest)`? Previously we would try to pull by `imageName` which was always present and not empty

src/transmitter/types.ts
dkontorovskyy: I tried to align it with previous view, but yeah seems totally legit to use `imageWithDigest`

src/scanner/images/index.ts
dkontorovskyy: You mean like `skopeoCopy("nginx")` e.g.?

dkontorovskyy: Plus, this flow happens way after `skopeoCopy` functionality.

src/scanner/index.ts
```diff
@@ -3,17 +3,25 @@ import { pullImages, removePulledImages, getImagesWithFileSystemPath, scanImages
...
 
 export async function processWorkload(workloadMetadata: IWorkload[]): Promise<void> {
   // every workload metadata references the same workload name, grab it from the first one
   const workloadName = workloadMetadata[0].name;
-  const allImages = workloadMetadata.map((meta) => meta.imageName);
-  logger.info({workloadName, imageCount: allImages.length}, 'queried workloads');
-  const uniqueImages = [...new Set<string>(allImages)];
+  const uniqueImages: { [key: string]: IScanImage } = workloadMetadata.reduce((accum, meta) => {
+    // example: "docker.io/library/redis@sha256:33ca074e6019b451235735772a9c3e7216f014aae8eb0580d7e94834fe23efb3"
+    const imageWithDigest = meta.imageId.substring(meta.imageId.lastIndexOf('/') + 1);
```
ivanstanev, line 13: Neat calculation, even if the image is something like `nginx@sha256:digest` it will end up as `-1 + 1 === 0`

src/scanner/index.ts
```diff
@@ -3,17 +3,25 @@ import { pullImages, removePulledImages, getImagesWithFileSystemPath, scanImages
...
-import { IPullableImage } from './images/types';
+import { IPullableImage, IScanImage } from './images/types';
 
 export async function processWorkload(workloadMetadata: IWorkload[]): Promise<void> {
   // every workload metadata references the same workload name, grab it from the first one
   const workloadName = workloadMetadata[0].name;
-  const allImages = workloadMetadata.map((meta) => meta.imageName);
-  logger.info({workloadName, imageCount: allImages.length}, 'queried workloads');
-  const uniqueImages = [...new Set<string>(allImages)];
+  const uniqueImages: { [key: string]: IScanImage } = workloadMetadata.reduce((accum, meta) => {
```
ivanstanev, line 11: Just to double check what this does, it groups all images by key if there are multiple of the same image listed in the workload, and then ensures we collect every digest? Is it possible the same image:digest may appear more than once?

ivanstanev (dismissed):
Nice!

src/scanner/index.ts
dkontorovskyy: @ivanstanev I believe it can't happen. Because `imageName` is actually `imageName:tag`. So `imageName:tag` will result to the same `imageName:digest`. And cos I'm reducing it, it will overwrite by key the same images.

ivanstanev (approved):

PR #587, opened 2020-10-27
fix: fallback to tag if digest missing

dkontorovskyy:

### What this does
Fallback to tag if digest missing

### DIFF
--- a/src/scanner/images/index.ts
+++ b/src/scanner/images/index.ts
@@ -10,13 +10,15 @@ export async function pullImages(images: IPullableImage[]): Promise<IPullableIma
   const pulledImages: IPullableImage[] = [];
 
   for (const image of images) {
-    const {imageWithDigest, fileSystemPath} = image;
+    const { imageName, imageWithDigest, fileSystemPath } = image;
     if (!fileSystemPath) {
       continue;
     }
 
     try {
-      await skopeoCopy(imageWithDigest, fileSystemPath);
+      // Scan image by digest if exists, other way fallback tag
+      const scanId = imageWithDigest ?? imageName;
+      await skopeoCopy(scanId, fileSystemPath);
       pulledImages.push(image);
     } catch (error) {
       logger.error({error, image: imageWithDigest}, 'failed to pull image');
@@ -93,11 +95,12 @@ export async function scanImages(images: IPullableImage[]): Promise<IScanResult[
       }
 
       const imageParts = getImageParts(imageName);
+      const imageDigest = imageWithDigest && getImageParts(imageWithDigest).imageDigest;
 
       result.imageMetadata = {
         image: imageParts.imageName,
         imageTag: imageParts.imageTag,
-        imageDigest: getImageParts(imageWithDigest).imageDigest,
+        imageDigest,
       };
 
       scannedImages.push({
--- a/src/scanner/images/index.ts
+++ b/src/scanner/images/index.ts
@@ -3,31 +3,31 @@ import * as plugin from 'snyk-docker-plugin';
 
 import logger = require('../../common/logger');
 import { pull as skopeoCopy, getDestinationForImage } from './skopeo';
-import { IPullableImage } from './types';
+import { IPullableImage, IScanImage } from './types';
 import { IStaticAnalysisOptions, StaticAnalysisImageType, IScanResult, IPluginOptions } from '../types';
 
 export async function pullImages(images: IPullableImage[]): Promise<IPullableImage[]> {
   const pulledImages: IPullableImage[] = [];
 
   for (const image of images) {
-    const {imageName, fileSystemPath} = image;
+    const {imageWithDigest, fileSystemPath} = image;
     if (!fileSystemPath) {
       continue;
     }
 
     try {
-      await skopeoCopy(imageName, fileSystemPath);
+      await skopeoCopy(imageWithDigest, fileSystemPath);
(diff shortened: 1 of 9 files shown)


src/scanner/images/index.ts
```diff
@@ -3,31 +3,33 @@ import * as plugin from 'snyk-docker-plugin';
...
+    const { imageName, imageWithDigest, fileSystemPath } = image;
     if (!fileSystemPath) {
       continue;
     }
 
     try {
-      await skopeoCopy(imageName, fileSystemPath);
+      // Scan image by digest if exists, other way fallback tag
+      const scanId = imageWithDigest ?? imageName;
+      await skopeoCopy(scanId, fileSystemPath);
```
dkontorovskyy, line 21: add fallback to tag

src/scanner/index.ts
```diff
@@ -35,6 +33,38 @@ export async function sendDeleteWorkloadRequest(workloadName: string, localWorkl
...
+    // example: For GCR "gcr.io/test-dummy/redis:latest"
+    // example: For ECR "291964488713.dkr.ecr.us-east-2.amazonaws.com/snyk/redis:latest"
+    // meta.imageName can be different depends on CR
+    const { imageName } = getImageParts(meta.imageName);
+    // meta.imageId can be different depends on CR
+    // example: For DCR "docker.io/library/redis@sha256:8e9f8546050da8aae393a41d65ad37166b4f0d8131d627a520c0f0451742e9d6"
+    // example: For GCR "sha256:8e9f8546050da8aae393a41d65ad37166b4f0d8131d627a520c0f0451742e9d6"
+    // example: For ECR "sha256:8e9f8546050da8aae393a41d65ad37166b4f0d8131d627a520c0f0451742e9d6"
+    let digest: string | undefined = undefined;
+    if (meta.imageId.lastIndexOf('@') > -1 || meta.imageId.startsWith('sha')) {
```
dkontorovskyy, line 53: only scan by digest if we sure it has it

ivanstanev: Shouldn't we try to remove all URI scheme prefixes first before looking for the SHA? So that we can handle things like `docker-pullable://`?

src/scanner/index.ts
```diff
@@ -35,6 +33,38 @@ export async function sendDeleteWorkloadRequest(workloadName: string, localWorkl
   await deleteWorkload(deletePayload);
 }
 
+export function getUniqueImages(workloadMetadata: IWorkload[]): IScanImage[] {
+  const uniqueImages: { [key: string]: IScanImage } = workloadMetadata.reduce((accum, meta) => {
+    logger.info({
+      workloadName: workloadMetadata[0].name,
+      name: meta.imageName,
+      id: meta.imageId
+    }, 'image metadata');
```
ivanstanev, line 42: Wouldn't this print super frequently?

test/unit/scanner/index.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        {
+            imageName: 'gcr.io/test-dummy/redis:latest',
+            imageId: 'sha256:8e9f8546050da8aae393a41d65ad37166b4f0d8131d627a520c0f0451742e9d6'
+        },
+        // 6. ECR
+        {
+            imageName: '291964488713.dkr.ecr.us-east-2.amazonaws.com/snyk/redis:latest',
+            imageId: 'sha256:8e9f8546050da8aae393a41d65ad37166b4f0d8131d627a520c0f0451742e9d6'
+        },
+    ];
```
ivanstanev, line 38: I think we need cases for the following:

```typescript
imageId: 'docker://sha256:abcd',
imageId: 'docker://name@sha256:abcd',
imageId: 'docker-pullable://name@sha256:abcd',
```

ivanstanev (commented):
Nice! 👏

src/scanner/index.ts
dkontorovskyy: I thought of giving this logger day or two, just to have sample data. WDYT?

test/unit/scanner/index.test.ts
dkontorovskyy: Cool. Will add test cases!

src/scanner/index.ts
dkontorovskyy: and then remove it

test/unit/scanner/index.test.ts
dkontorovskyy: Still, the first case is not supported for now. So I would skip it. When we get some sample data from the logger, I might add this. WDYT?

src/scanner/index.ts
dkontorovskyy: Prefer doing this out of data from logger. Cos we still kinda blind what `imageId` can be in all variations.

ivanstanev: We'll test it in the environments we have? We have GKE, EKS, OpenShift but nothing on AKS or any other managed service 🤔 But I guess it's a good start

src/scanner/index.ts
ivanstanev: Sure!

test/unit/scanner/index.test.ts
ivanstanev: Sure thing, as long as we can fallback to the old way of pulling by name/tag 🙏

ivanstanev (approved):

PR #602, opened 2020-11-18
Feat/podspec from pods

ivanstanev:

### What this does

PodSpec is now read from Pods instead of the workloads.

This is a workaround for cases where the workload is a Custom Resource, which snyk-monitor today does not support and cannot read them.
There are situations where we will misreport the securityContext because we tried to read it from the wrong workload.

The Pod's spec is an identical copy to the one defined in the workload, so we can use it for collecting the workload information that we need - this way we can track workload security configuration even if the Pods are managed by a Custom Resource.

### More information

- [ZenDesk #7587](https://snyk.zendesk.com/agent/tickets/7587)

**Working example**
Normally if there is a Deployment we will have the following chain of resources:
```
Deployment -> ReplicaSet -> Pod
```
snyk-monitor inspects the Pod and then starts walking up the parent references (ReplicaSet and then Deployment) to try and understand the actual workload. Since the Deployment is last, we choose it to be the workload. We then read the Deployment's configuration (resources, securityContext, etc.).

If we have a Custom Resource, there can be a chain like this:
```
CustomResource -> ReplicaSet -> Pod
```
snyk-monitor does not support custom resources so when inspecting the parents it will stop at the ReplicaSet. The problem is that the ReplicaSet configuration (resources, securityContext) may be empty/wrong/incomplete so we would be reading the _wrong_ data.

### DIFF
--- a/src/supervisor/metadata-extractor.ts
+++ b/src/supervisor/metadata-extractor.ts
@@ -1,4 +1,4 @@
-import { V1OwnerReference, V1Pod, V1Container, V1ContainerStatus } from '@kubernetes/client-node';
+import { V1OwnerReference, V1Pod, V1Container, V1ContainerStatus, V1PodSpec } from '@kubernetes/client-node';
 import { IWorkload, ILocalWorkloadLocator } from '../transmitter/types';
 import { currentClusterName } from './cluster';
 import { IKubeObjectMetadata } from './types';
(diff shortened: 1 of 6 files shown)


agranado2k (approved):
LGTM! 🎉

dkontorovskyy (approved):
LGTM

PR #608, opened 2020-11-23
chore: add missing npm script to fix `npm t`

michael-c-snyk:
`test:integration` was missing; this redirects it to most useful fast integration test for convenience of testing via `npm t`. This is a benefit to local dev, since `npm t` isn't currently used in CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "main": "dist/index.js",
   "scripts": {
     "pretest": "./scripts/docker/build-image.sh",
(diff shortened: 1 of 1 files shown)


ivanstanev (commented):
Not sure we want to ever test with all suites sequentially, this means an 1+ hour long test run! I think we should test only one combination - `test:integration:kind:helm`

ivanstanev (approved):

PR #611, opened 2020-11-23
feat:support storage in PVC

michael-c-snyk:
[RUN-1250] Update helm templates to allow for a `pvc` value. When set, this replaces emptyDir with a PVC, backed by a PV whose provsiioning is handled separately

- [ ] Documentation written [ℹ︎](https://github.com/snyk/general/wiki/Documentation) (inline docs done, rest to follow)

### What this does

Allows users to store temporary files in custom storage via a PVC, rather than needing an empty node.

### Notes for the reviewer

see slack channel # run-1250-demo

### More information

[RUN-1250]

### Screenshots

see slack channel # run-1250-demo

[RUN-1250]: https://snyksec.atlassian.net/browse/RUN-1250
[RUN-1250]: https://snyksec.atlassian.net/browse/RUN-1250

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -20,6 +20,13 @@ spec:
     spec:
       serviceAccountName: {{ include "snyk-monitor.name" . }}
       restartPolicy: Always
+      initContainers:
+        - name: volume-permissions
+          image: "{{ .Values.initContainerImage.repository }}:{{ .Values.initContainerImage.tag }}"
+          command : ['sh', '-c', 'chmod -R 777 /var/tmp']
+          volumeMounts:
+            - name: temporary-storage
+              mountPath: "/var/tmp"
       containers:
         - name: {{ include "snyk-monitor.name" . }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
@@ -83,8 +90,13 @@ spec:
               - key: dockercfg.json
                 path: config.json
         - name: temporary-storage
+          {{- if .Values.pvc.enabled }}
+          persistentVolumeClaim:
+            claimName: {{ .Values.pvc.name }}
+          {{- else }}
           emptyDir:
             sizeLimit: {{ .Values.temporaryStorageSize }}
+          {{- end }}
         - name: ssl-certs
           configMap:
             name: {{ .Values.certsConfigMap }}
--- a/snyk-monitor/templates/pvc.yaml
+++ b/snyk-monitor/templates/pvc.yaml
@@ -1,0 +1,13 @@
+{{ if .Values.pvc.enabled }}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ .Values.pvc.name }}
+spec:
+  accessModes:
+    - ReadWriteOnce
+  volumeMode: Filesystem
+  resources:
+    requests:
+      storage: {{ .Values.temporaryStorageSize }}
+{{ end }}
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -22,14 +22,24 @@ image:
   tag: IMAGE_TAG_OVERRIDE_WHEN_PUBLISHING
   pullPolicy: Always
 
+# If deploying in an air-gapped environment that can't pull from DockerHub, override the initContainer's image here for one that is accessible to your environment.
+initContainerImage:
+  repository: busybox
+  tag: latest
+
 # The snyk-monitor requires knowing the cluster name so that it can organise
 # scanned workloads. The Kubernetes API does not provide an API to query this.
 # Set the name of the cluster, otherwise the snyk-monitor will set this to a default value.
 clusterName: ""
 
 # The snyk-monitor requires disk storage to temporarily pull container images and to scan them for vulnerabilities.
-# This value controls how much disk storage _at most_ may be allocated for the snyk-monitor. The snyk-monitor mounts an emptyDir for storage.
-temporaryStorageSize: 50Gi
+# This value controls how much disk storage _at most_ may be allocated for the snyk-monitor. Unless overridden by the `pvc` value, the snyk-monitor mounts an emptyDir for storage.
+temporaryStorageSize: 50Gi #  Applies to PVC too
+
+# Change to true to use a PVC instead of emptyDir for local storage
+pvc:
+  enabled: false
+  name: 'snyk-monitor-pvc'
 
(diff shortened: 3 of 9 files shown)


snyk-monitor/templates/deployment.yaml
```diff
@@ -83,8 +83,13 @@ spec:
               - key: dockercfg.json
                 path: config.json
         - name: temporary-storage
+          {{ if .Values.pvc }}
```
ivanstanev, line 86: Should this be `.Values.pvc.enabled`?

snyk-monitor/templates/pvc.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if .Values.pvc }}
```
agranado2k, line 1: Just don't forget to be consistent with the previous @ivanstanev comment `.Values.pvc.enabled`

agranado2k (commented):
Great work!
I added a reminder.

snyk-monitor/templates/pvc.yaml
michael-c-snyk: Good spot, thanks!

snyk-monitor/templates/deployment.yaml
michael-c-snyk: It definitely should =D

snyk-monitor/values.yaml
```diff
@@ -28,8 +28,13 @@ image:
 clusterName: ""
 
 # The snyk-monitor requires disk storage to temporarily pull container images and to scan them for vulnerabilities.
-# This value controls how much disk storage _at most_ may be allocated for the snyk-monitor. The snyk-monitor mounts an emptyDir for storage.
-temporaryStorageSize: 50Gi
+# This value controls how much disk storage _at most_ may be allocated for the snyk-monitor. The snyk-monitor mounts an emptyDir for storage. # TODO updateme
```
ivanstanev, line 31: There is a small `# TODO updateme` in here! 🙏

michael-c-snyk: argh

michael-c-snyk: Good spot! fixed

ivanstanev (dismissed):
Looks good! I guess we should hold off merging until we test on `dev`?

snyk-monitor/templates/deployment.yaml
```diff
@@ -20,6 +20,13 @@ spec:
     spec:
       serviceAccountName: {{ include "snyk-monitor.name" . }}
       restartPolicy: Always
+      initContainers:
+        - name: volume-permissions
+          image: busybox
+          command : ['sh', '-c', 'chmod -R 777 /var/tmp']
```
michael-c-snyk, line 26: 777 is really lax, but nothing less restrictive seems to work, because we're not guaranteed to know the user/group id until everything is deployed. Can anyone suggest a better alternative?

snyk-monitor/templates/pvc.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if .Values.pvc.enabled }}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ .Values.pvc.name }}
+spec:
+  accessModes:
+    - ReadWriteOnce
```
michael-c-snyk, line 8: As mentioned in Slack, there's a tradeoff here. To support a large number of storage drivers (including EBS which our customer uses), we specify ReadWriteOnce, which only supports access from a single pod. This _may_ make it harder to horizontally scale snyk-monitor to multiple pods in future

src/index.ts
```diff
@@ -33,6 +36,16 @@ process.on('unhandledRejection', (reason) => {
   }
 });
 
+function cleanUpTempStorage() {
+  const { IMAGE_STORAGE_ROOT } = config;
+  try {
+    emptyDirSync(IMAGE_STORAGE_ROOT);
```
michael-c-snyk, line 42: Customer has complained that sometimes storage isnt emptied properly. This is likely to be an even bigger concern if we inherited a PV used elsewhere, so we now ensure it's empty upon startup

Shesekino:
really interesting to see the usage of initcontainer here!
how did you decide to do that over running the command through the Snyk Monitor when it starts?

Shesekino:
sounds brilliant :)

snyk-monitor/templates/deployment.yaml
ivanstanev: Do we need the `execute` bit? We usually only read and write files to that directory

snyk-monitor/templates/deployment.yaml
```diff
@@ -20,6 +20,13 @@ spec:
     spec:
       serviceAccountName: {{ include "snyk-monitor.name" . }}
       restartPolicy: Always
+      initContainers:
+        - name: volume-permissions
+          image: busybox
```
ivanstanev, line 25: This must be configurable actually, otherwise it won't work for on-prem air-gapped deployments (meaning without internet access, with some local container Registry where users have only approved images that they allow on the cluster) 🙏

michael-c-snyk: Ah, I hadn't thought of that - great point! Will add and document an override 👍

snyk-monitor/templates/deployment.yaml
michael-c-snyk: Good point. It has execute by default, but you're right that I dont think it's needed - will test and update 👍

michael-c-snyk: Just tested and even chmod 776 without  the `execute` permissions seems to bring the "Permission denied" error back - looks like it is needed for some reason 🤔

snyk-monitor/templates/deployment.yaml
michael-c-snyk: I guess this means that the initContainer should mount the `docker-config` volume too, if I'm right in thinking that this is where it gets private registry credentials from?

ivanstanev: I don't think so, the `docker-config` volume is for `kubernetes-monitor` to be able to pull and analyze images from private registries. This image here is a request from Kubernetes itself to deploy it for us

michael-c-snyk: Aha ok. So if customer wants to use a local container registry to get an alternative image to busybox, do they need any new way to send credentials, or is this already taken care of somewhere?

ivanstanev: They will configure it themselves yep! 🙏

ivanstanev (approved):

PR #618, opened 2020-11-30
feat: memory optimization from depGraph lib

dkontorovskyy:

### What this does
Bring memory optimization from depGraph lib

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3729,11 +3729,11 @@
       }
     },
     "snyk-docker-plugin": {
-      "version": "4.12.0",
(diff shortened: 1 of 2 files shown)


ivanstanev (commented):
I really prefer to pin to specific versions, we've had cases where we cause accidental breaking changes... One example is my fault with `rpm-parser` which then propagated errors all the way to the Snyk CLI

dkontorovskyy:
@ivanstanev we are bringing only the latest `patch` version, which are `fix` commits and not breaking functionality

ivanstanev (approved):

PR #620, opened 2020-12-01
feat: Add OpenShift configuration for supporting PVC storage

michael-c-snyk:
feat: Add OpenShift configuration for supporting PVC storage

This was previously exposed via Helm, now making it available for the Operator installation as well.

This change allows both the PVC storage to be enabled and the initContainer image to be configured.

### DIFF
--- a/snyk-operator/deploy/olm-catalog/snyk-operator/0.0.0/snyk-operator.v0.0.0.clusterserviceversion.yaml
+++ b/snyk-operator/deploy/olm-catalog/snyk-operator/0.0.0/snyk-operator.v0.0.0.clusterserviceversion.yaml
@@ -20,7 +20,15 @@ metadata:
             "integrationApi": "",
             "monitorSecrets": "snyk-monitor",
             "scope": "Cluster",
-            "temporaryStorageSize": "50Gi"
+            "temporaryStorageSize": "50Gi",
+            "pvc": {
+              "enabled": false,
+              "name": "snyk-monitor-pvc"
+            },
+            "initContainerImage": {
+              "repository": "busybox",
+              "tag": "latest"
+            }
           }
         }
       ]
@@ -50,6 +58,9 @@ spec:
             version: v1
           - kind: Secret
             name: snyk-monitor
+            version: v1
+          - kind: PersistentVolumeClaim
+            name: snyk-monitor-pvc
             version: v1
         specDescriptors:
           - description: >-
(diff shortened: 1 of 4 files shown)


snyk-operator/deploy/olm-catalog/snyk-operator/0.0.0/snyk-operator.v0.0.0.clusterserviceversion.yaml
```diff
@@ -51,6 +59,9 @@ spec:
           - kind: Secret
             name: snyk-monitor
             version: v1
+          - kind: PersistentVolumeClaim
+            name: snyk-monitor-pvc
```
michael-c-snyk, line 63: @ivanstanev Is this OK given that the pvc's name could be overridden by a Value?

ivanstanev: I am not sure, we could test it with a PVC named differently than this! But I think these fields here are going to be shown in the UI, perhaps not critical and won't block deployment

snyk-operator/deploy/olm-catalog/snyk-operator/0.0.0/snyk-operator.v0.0.0.clusterserviceversion.yaml
```diff
@@ -103,6 +114,30 @@ spec:
             path: image.tag
             x-descriptors:
             - 'urn:alm:descriptor:text'
+          - description: >-
+              True to use a PVC for temporary storage, false to use emptyDir
+            displayName: PVC enabled
+            path: pvc.enabled
+            x-descriptors:
+              - 'urn:alm:descriptor:com.tectonic.ui:booleanSwitch'
```
ivanstanev, line 122: Nice!

snyk-operator/deploy/olm-catalog/snyk-operator/0.0.0/snyk-operator.v0.0.0.clusterserviceversion.yaml
```diff
@@ -103,6 +114,30 @@ spec:
             path: image.tag
             x-descriptors:
             - 'urn:alm:descriptor:text'
+          - description: >-
+              True to use a PVC for temporary storage, false to use emptyDir
```
ivanstanev, line 118: Just to note we should probably add a full stop to all of these descriptions since they're shown in the UI

test/fixtures/operator/custom-resource.yaml
```diff
@@ -5,3 +5,10 @@ metadata:
   namespace: snyk-monitor
 spec:
   integrationApi: https://kubernetes-upstream.dev.snyk.io
+  clusterName: My dev clusterName
+  pvc:
+    enabled: false
+    name: snyk-monitor-pvc
+  initContainerImage:
+    repository: busybox
+    tag: latest
```
ivanstanev, line 14: Should we enable this on the OpenShift cluster? If we do, I think we should reduce the disk size to something like 10-15 GB because the cluster was provisioned with a pretty small disk I think 😓 

Also we don't need to list clusterName and initContainerImage, they will use the default values

shaimendel (approved):

PR #635, opened 2021-01-10
chore: install helm from package manager

dkontorovskyy:

### What this does
Install Helm during the release process and stop being dependant on a very outdated version of helm that was mysteriously living in `gh-pages` branch as a committed to SCM binary

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,16 @@
 commands:
+    install_helm:
+        description: Install requests library
+        steps:
+            - run:
+                command: |
+                    export DEBIAN_FRONTEND=noninteractive
+                    curl https://baltocdn.com/helm/signing.asc | sudo apt-key add -
+                    sudo apt-get install -y apt-transport-https
+                    echo "deb https://baltocdn.com/helm/stable/debian/ all main" | sudo tee /etc/apt/sources.list.d/helm-stable-debian.list
+                    sudo apt-get update -qq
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -1,4 +1,16 @@
...
+        description: Install requests library
+        steps:
+            - run:
+                command: |
+                    export DEBIAN_FRONTEND=noninteractive
+                    curl https://baltocdn.com/helm/signing.asc | sudo apt-key add -
+                    sudo apt-get install -y apt-transport-https
+                    echo "deb https://baltocdn.com/helm/stable/debian/ all main" | sudo tee /etc/apt/sources.list.d/helm-stable-debian.list
+                    sudo apt-get update -qq
+                    sudo apt-get install -y helm
```
ivanstanev, line 12: Nit: I think we could do all of this as a single command, the docs say we can install through Snap: `sudo snap install helm --classic`

ivanstanev (approved):

.circleci/config.yml
dkontorovskyy: Snap doesn't appear to be present in circleci pre-built images, can you point me to docs 🙏 ?

ivanstanev: Ah interesting... I don't know enough about it, perhaps it depends on the Ubuntu version or it's controlled by some env var or flag. Let's leave it as is, your approach is also the official way

PR #637, opened 2021-01-11
chore: remove call to Helm's deprecated `init` API

dkontorovskyy:

### What this does
Remove call to Helm's deprecated `init` API

### DIFF
--- a/scripts/publish-gh-pages.sh
+++ b/scripts/publish-gh-pages.sh
@@ -28,7 +28,6 @@ sed -i "s/IMAGE_TAG_OVERRIDE_WHEN_PUBLISHING/${NEW_TAG}/g" ./snyk-monitor/values
 sed -i "s/IMAGE_TAG_OVERRIDE_WHEN_PUBLISHING/${NEW_TAG}/g" ./snyk-monitor-deployment.yaml
 
 echo building new helm release
-helm init --client-only
 helm package snyk-monitor --version ${NEW_TAG}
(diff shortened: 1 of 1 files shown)


scripts/publish-gh-pages.sh
```diff
@@ -28,7 +28,7 @@ sed -i "s/IMAGE_TAG_OVERRIDE_WHEN_PUBLISHING/${NEW_TAG}/g" ./snyk-monitor/values
 sed -i "s/IMAGE_TAG_OVERRIDE_WHEN_PUBLISHING/${NEW_TAG}/g" ./snyk-monitor-deployment.yaml
 
 echo building new helm release
-helm init --client-only
+helm lint snyk-monitor
```
shaimendel, line 31: Don't we still need the helm init command after it?

ivanstanev: Kinda seems like a v2 / Tiller thing, init doesn't exist in v3 Helm

dkontorovskyy: It's deprecated from helm v3

image

shaimendel: another Q - what does `helm lint` does? I wonder if it is better to fail the build if you didn't manually lint the helm before you commit (like in the `test` script). otherwise we might publish a different version of the helm from what we have in staging/master

shaimendel (approved):

PR #645, opened 2021-01-26
feat: add quotes for .Values.skip_k8s_jobs, add tolerations

nicholasklem:
Untested changes:
- quote .Values.skip_k8s_jobs
- add support for tolerations in the helm chart.

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -122,6 +122,7 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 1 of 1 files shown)


snyk-monitor/templates/deployment.yaml
```diff
@@ -122,3 +122,6 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
jmickey, line 127:
```suggestion
      {{- with .Values.tolerations }}
      tolerations:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

jmickey (changes requested):
Hi Nicholas,

Thanks so much for taking the time to make this PR, we really appreciate it. I just have slight change, as there appears to be a missing directive for the tolerations.

Additionally, would you mind amending the commit message to prepend `feat:` so that our semantic release processes are triggererd correctly?

PR #653, opened 2021-02-07
feat: restart k8s informers on error event (e.g idle connection)

dkontorovskyy:

### What this does
Restart k8s informers on error event (e.g idle connection)

### DIFF
--- a/src/supervisor/watchers/handlers/index.ts
+++ b/src/supervisor/watchers/handlers/index.ts
@@ -1,4 +1,4 @@
-import { makeInformer, ADD, DELETE, UPDATE, KubernetesObject } from '@kubernetes/client-node';
+import { makeInformer, ADD, DELETE, ERROR, UPDATE, KubernetesObject } from '@kubernetes/client-node';
 import { logger } from '../../../common/logger';
 import { WorkloadKind } from '../../types';
 import { podWatchHandler, podDeletedHandler } from './pod';
@@ -12,6 +12,7 @@ import { statefulSetWatchHandler } from './stateful-set';
 import { k8sApi, kubeConfig } from '../../cluster';
 import * as kubernetesApiWrappers from '../../kuberenetes-api-wrappers';
 import { IWorkloadWatchMetadata, FALSY_WORKLOAD_NAME_MARKER } from './types';
+import { ECONNRESET_ERROR_CODE } from '../types';
 
 /**
  * This map is used in combination with the kubernetes-client Informer API
@@ -108,6 +109,20 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
   };
 
   const informer = makeInformer<KubernetesObject>(kubeConfig, namespacedEndpoint, loggedListMethod);
+
+  informer.on(ERROR, (err) => {
+    // Types from client library insists that callback is of type KubernetesObject
+    if ((err as any).code === ECONNRESET_ERROR_CODE) {
+      logger.debug(`informer ${ECONNRESET_ERROR_CODE} occurred, restarting informer`);
+
+      // Restart informer after 1sec
+      setTimeout(() => {
+        informer.start();
+      }, 1000);
+    } else {
+      logger.error({ err }, 'unexpected informer error event occurred');
+    }
+  });
 
   for (const informerVerb of Object.keys(workloadMetadata.handlers)) {
     informer.on(informerVerb, async (watchedWorkload) => {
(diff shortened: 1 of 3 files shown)


src/supervisor/watchers/handlers/index.ts
```diff
@@ -109,6 +109,15 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
 
   const informer = makeInformer<KubernetesObject>(kubeConfig, namespacedEndpoint, loggedListMethod);
 
+  informer.on(ERROR, (err) => {
+    logger.error({ namespace, workloadKind, err }, 'informer error event occurred');
```
ivanstanev, line 113: Is it possible to distinguish "connection interrupted" because of, for example, AKS timeout vs. other errors? In the first case we can treat it as a warn (even info), and the latter as an actual error. We could then add a max retry attempt on the error case and end the program, as retrying is not always the best case (e.g. just spinning in an error loop that has no resolution).

src/supervisor/watchers/handlers/index.ts
```diff
@@ -109,6 +109,15 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
...
+
+    // Restart informer after 5sec
+    setTimeout(() => {
+      informer.start();
+    }, 1000);
```
ivanstanev, line 118: Should we update the comment to 1sec or update the timeout to 5000?

ivanstanev (commented):
Incredible find 👏 Well done!

Added some minor questions

src/supervisor/watchers/handlers/index.ts
dkontorovskyy: @ivanstanev I'm happy to narrow it down to `ECONNRESET`. And for other errors just log for now. WDYT?
```
{"name":"kubernetes-monitor","hostname":"snyk-monitor-779d797fc7-zddtk","pid":6,"level":50,"namespace":"snyk-monitor","workloadKind":"StatefulSet","err":{"message":"read ECONNRESET","name":"Error","stack":"Error: read ECONNRESET\n    at TLSWrap.onStreamRead (internal/stream_base_commons.js:209:20)","code":"ECONNRESET"},"msg":"informer error event occurred","time":"2021-02-07T19:18:44.991Z","v":0}
```

ivanstanev: Sounds good! :)

src/supervisor/watchers/handlers/index.ts
```diff
@@ -109,6 +110,20 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
...
+    // Types from client library insists that callback is of type KubernetesObject
+    if ((err as any).code === ECONNRESET_ERROR_CODE) {
+      logger.debug('informer ECONNRESET occurred, restarting informer');
+
+      // Restart informer after 1sec
+      setTimeout(() => {
+        informer.start();
+      }, 1000);
+    } else {
+      logger.error({ err }, 'unexpected informer error event occurred');
```
ivanstanev, line 123: Should we also throw to kill the program? Or we don't do it so that we don't lose the logs?

src/supervisor/watchers/handlers/index.ts
```diff
@@ -109,6 +110,20 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
 
   const informer = makeInformer<KubernetesObject>(kubeConfig, namespacedEndpoint, loggedListMethod);
 
+  informer.on(ERROR, (err) => {
+    // Types from client library insists that callback is of type KubernetesObject
+    if ((err as any).code === ECONNRESET_ERROR_CODE) {
```
ivanstanev, line 115: I think I'd be fine comparing directly to `'ECONNRESET'`, up to you! We actually compare to these network codes in another place too (`retryKubernetesRequest()` or whatever the name of the function).

ivanstanev (approved):

src/supervisor/watchers/handlers/index.ts
dkontorovskyy: I think this is too specific a place and don't involve retries like the retry function as well as that function has very specific cases to querying k8s API

src/supervisor/watchers/handlers/index.ts
dkontorovskyy: We ideally shouldn't get to here, so it would be nice to have logs if smth happens, so we can act on that.
You can always have logs from previous pod by doing `kc logs <pod-id> --previos`

PR #656, opened 2021-02-09
chore: pass process.env when calling child_process.exec()

ivanstanev:

### What this does

It seems like there is an issue with Jest where changes to process.env are not persisted through calls to child_process.exec(). This causes env vars like KUBECONFIG to be lost and to fail our tests. For now use a wrapper to explicitly pass process.env as a workaround.

### More information

- [Investigation notes](https://snyk.slack.com/archives/CLW30N31V/p1612860708027800?thread_ts=1612856703.026500&cid=CLW30N31V)

### DIFF
--- a/test/helpers/exec.ts
+++ b/test/helpers/exec.ts
@@ -1,0 +1,13 @@
+import { exec, PromiseResult } from 'child-process-promise';
+
+/**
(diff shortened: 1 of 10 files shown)


dkontorovskyy (approved):
LGTM

PR #657, opened 2021-02-09
Add Helm option and Scanning Namespace

rmkanda:
README Updated with how to scan a namespace alone.


ivanstanev:
Hey @rmkanda, thanks so much for the change! The instructions are currently present in here: https://github.com/snyk/kubernetes-monitor/tree/staging/snyk-monitor, I guess it's not very obvious, there is another README as part of this Helm chart.

Do you think it may be better to link to the existing README?

And another small request, apologies that it's not in the contributing guidelines, but could you prefix your commit with `chore:` as we use semantic-release to track our commits/releases?

ivanstanev:
Hey @rmkanda just checking again if you had the chance to look at the previous suggestion?

PR #664, opened 2021-02-20
[RUN-1388] Throttle In-Flight Requests to Homebase

jmickey:

### What this does

This PR implements a queue that wraps calls to the `retryRequest` function. Functionally this should limit the number of concurrent requests to `homebase` to the size of the queue - in this case I have limited it to 2.

The effect of this still requires some additional testing. 

### Notes for the reviewer

### More information

- [Jira ticket SC-0000](https://snyksec.atlassian.net/browse/RUN-1388)

### Screenshots


src/transmitter/index.ts
```diff
@@ -12,16 +12,40 @@ import {
...
+    IWorkloadMetadataPayload |
+    IDeleteWorkloadPayload;
+}
 
 const upstreamUrl = config.INTEGRATION_API || config.DEFAULT_KUBERNETES_UPSTREAM_URL;
 
+// Async queue wraps around the call to retryRequest in order to limit
+// the number of requests in flight to Homebase at any one time.
+const reqQueue = queue(async function(req: HomebaseRequest) {
+  return await retryRequest(req.method, req.url, req.payload);
```
ivanstanev, line 32: Might need to wrap in try-catch to avoid unhandled promise rejections

ivanstanev (commented):
Great stuff! 👏 

We also may want to remove the queue that protects "sending workload metadata to Homebase" defined here: https://github.com/snyk/kubernetes-monitor/blob/2f50c1ca50bacc56047f90ab57911a1188632fd6/src/supervisor/watchers/handlers/pod.ts#L42-L47 It had a concurrency of 10 and only applied to sending workload meta, I like that this approach handles _everything_ now 👏

src/transmitter/index.ts
jmickey: The call to push is awaited and wrapped in try-catch. I'll need to check but that should avoid unhandled promise rejections:
https://github.com/snyk/kubernetes-monitor/pull/664/files#diff-d40843739012d28af7cb08fc87c96ded838baa28d5a6c617d41e75d9b5402d54R72

ivanstanev: Doesn't that line just enqueue the request? I think the callback here executes the request asynchronously with some sort of worker, but unfortunately `retryRequest()` can throw

jmickey: So, if I'm reading this page correctly: https://caolan.github.io/async/v3/global.html then an error should bubble up to the call to `push`.

I'll install it into a playground project and test it to verify though.

jmickey: I can confirm based on testing that this works as I had hoped.

src/transmitter/index.ts
```diff
@@ -12,16 +12,40 @@ import {
   IDependencyGraphPayload,
 } from './types';
 import { getProxyAgent } from './proxy';
+import { queue } from 'async';
```
ivanstanev, line 15: Nit: should be at the top with other `node_modules` imports

src/transmitter/index.ts
```diff
@@ -12,16 +12,40 @@ import {
...
     try {
-      const {response, attempt} = await retryRequest('post', `${upstreamUrl}/api/v1/dependency-graph`, payload);
+      const request: HomebaseRequest = {
+        method: 'post',
+        url: `${upstreamUrl}/api/v1/dependency-graph`,
+        payload: payload,
+      };
+
+      const { response, attempt } = await reqQueue.pushAsync(request);
+      // const {response, attempt} = await retryRequest('post', `${upstreamUrl}/api/v1/dependency-graph`, payload);
```
ivanstanev, line 48: Nit: can delete the comment (and in the other places where it happens)

src/transmitter/index.ts
dkontorovskyy, line 15: nit: move library imports to the top of the file, please 🙏

jmickey: I refuse.


jk. Will do that now

test/unit/supervisor/pod-watch-handler-caches.test.ts
```diff
@@ -47,14 +47,14 @@ tap.test('image and workload image cache', async (t) => {
   const podObject: V1Pod = YAML.parse(podFixture);
   await pod.podWatchHandler(podObject);
   await sleep(500);
-  t.equals(pushCallCount, 2, 'pushed data to send');
+  t.equals(pushCallCount, 1, 'pushed data to send');
```
ivanstanev, line 50: Wow I had no idea this test file exists 😆

src/transmitter/index.ts
```diff
@@ -39,7 +63,14 @@ export async function sendScanResults(payloads: ScanResultsPayload[]): Promise<b
...
     try {
-      const {response, attempt} = await retryRequest('post', `${upstreamUrl}/api/v1/scan-results`, payload);
+      const request: HomebaseRequest = {
+        method: 'post',
+        url: `${upstreamUrl}/api/v1/scan-results`,
+        payload: payload,
+      };
+
+      const { response, attempt } = await reqQueue.pushAsync(request);
+      // const {response, attempt} = await retryRequest('post', `${upstreamUrl}/api/v1/scan-results`, payload);
```
dkontorovskyy, line 73: Left-overs?

src/transmitter/index.ts
dkontorovskyy, line 48: Left-overs?

ivanstanev (approved):
Small nits to fix but other than that 🚀

src/transmitter/index.ts
```diff
@@ -71,7 +109,14 @@ export async function sendWorkloadMetadata(payload: IWorkloadMetadataPayload): P
...
   try {
-    const {response, attempt} = await retryRequest('delete', `${upstreamUrl}/api/v1/workload`, payload);
+    const request: HomebaseRequest = {
+      method: 'delete',
+      url: `${upstreamUrl}/api/v1/workload`,
+      payload: payload,
+    };
+
+    const { response, attempt } = await reqQueue.pushAsync(request);
+    // const {response, attempt} = await retryRequest('delete', `${upstreamUrl}/api/v1/workload`, payload);
```
dkontorovskyy, line 119: Left-overs?

test/unit/supervisor/pod-watch-handler-caches.test.ts
dkontorovskyy, line 50: Can you Jestify the tests 🙏 
Based on our discussion, you touched it, you jest it :)

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs lts-erbium
```
dkontorovskyy, line 1: Why we need this?

jmickey: https://github.com/asdf-vm/asdf

config.default.json
```diff
@@ -12,7 +12,7 @@
     "MAX_AGE_MS": 60000
   },
   "WORKLOADS_TO_SCAN_QUEUE_WORKER_COUNT": 10,
-  "METADATA_TO_SEND_QUEUE_WORKER_COUNT": 10,
+  "REQUEST_QUEUE_LENGTH": 2,
```
dkontorovskyy, line 15: Why it's 2?

jmickey: Assuming a response time of 200-300ms then 2 concurrent requests should average out to about 400-500 requests per minute. I think that is a decent starting point. We can tweak as we go

test/unit/supervisor/pod-watch-handler-caches.test.ts
jmickey: Updated it.

I considered just deleting these tests because they're kinda pointless but whatever

ivanstanev (approved):

PR #665, opened 2021-02-21
feat: support excluding namespaces from scanning

jmickey:

### What this does

Adds support for customers to specify a list of namespaces they would like to exclude from snyk-monitor's scanning. 

### Notes for the reviewer

### More information

- [Jira ticket SC-0000](https://snyksec.atlassian.net/browse/SC-0000)
- [Link to documentation](https://github.com/Snyk/registry/wiki/)

### Screenshots

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -1,14 +1,28 @@
 import * as uuidv4 from 'uuid/v4';
 import { loadConfig } from 'snyk-config';
+import * as fs from 'fs';
 
 const config: Record<string, any> = loadConfig(__dirname + '/../..', {
   secretConfig: process.env.CONFIG_SECRET_FILE,
 });
 
+const namespacesFilePath = '/etc/config/namespaces';
+
+function loadExcludedNamespaces(): string[] | null {
+  try {
+    const data = fs.readFileSync(namespacesFilePath, 'UTF-8');
+    const namespaces: string[] = data.split(/\r?\n/);
+    return namespaces;
+  } catch (err) {
+    return null;
+  }
+}
+
 config.AGENT_ID = uuidv4();
 config.INTEGRATION_ID = config.INTEGRATION_ID.trim();
 config.CLUSTER_NAME = config.CLUSTER_NAME || 'Default cluster';
 config.IMAGE_STORAGE_ROOT = '/var/tmp';
+config.EXCLUDED_NAMESPACES = loadExcludedNamespaces();
 
 /**
  * Important: we delete the following env vars because we don't want to proxy requests to the Kubernetes API server.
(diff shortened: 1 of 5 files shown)


src/common/config.ts
```diff
@@ -1,14 +1,28 @@
...
 const config: Record<string, any> = loadConfig(__dirname + '/../..', {
   secretConfig: process.env.CONFIG_SECRET_FILE,
 });
 
+const namespacesFilePath = '/etc/config/namespaces';
+
+function loadExcludedNamespaces(): string[] | null {
+  try {
+    const data = fs.readFileSync(namespacesFilePath, 'UTF-8');
+    const namespaces: string[] = data.split(/\r?\n/);
```
dkontorovskyy, line 14:
```suggestion
    const namespaces: string[] = data.split(/\r?\n/).filter(Boolean);
```
It will filter out any empty lines if there will be case like that

src/common/config.ts
```diff
@@ -1,14 +1,28 @@
 import * as uuidv4 from 'uuid/v4';
 import { loadConfig } from 'snyk-config';
+import * as fs from 'fs';
 
 const config: Record<string, any> = loadConfig(__dirname + '/../..', {
   secretConfig: process.env.CONFIG_SECRET_FILE,
 });
 
+const namespacesFilePath = '/etc/config/namespaces';
```
dkontorovskyy, line 9: Where is Helm charts you define this localtion?

dkontorovskyy (commented):
I 💯  😍 this PR 

Couple of questions and suggestions:
1. Would you mind adding some tests
2. Do we also need to update helm yaml im the root folder?

src/common/config.ts
jmickey: https://github.com/snyk/kubernetes-monitor/pull/665/files#diff-4d1b7a6d5d6f40fba7fa252f965e47092c24f6c7715d9be94adc3b320b303b12R54-R57

ivanstanev: `snyk-monitor/templates/configmap.yaml`

src/common/config.ts
```diff
@@ -1,14 +1,28 @@
...
+
+function loadExcludedNamespaces(): string[] | null {
+  try {
+    const data = fs.readFileSync(namespacesFilePath, 'UTF-8');
+    const namespaces: string[] = data.split(/\r?\n/);
+    return namespaces;
+  } catch (err) {
```
ivanstanev, line 16: Should we log in case users have made an error/typo and we need to debug it?

ivanstanev (commented):
Looks good! Note in the future we may also have a mechanism to "monitor workloads only in these listed namespaces" so we should consider this approach in the context of future changes. Also in the context of using Rego for other policies work, wonder if we could use the same mechanism everywhere 🤷

ivanstanev:
Closing so we stop getting alerts

PR #669, opened 2021-02-22
feat: load user-defined Rego rules for workload auto-import

ivanstanev:

### What this does

The Rego rules are loaded by snyk-monitor at the start of the application, the rules are verified, and finally they are transmitted to upstream.
We want the rules to be compiled and processed client-side but any choice whether workloads should be imported or deleted happens in the upstream.
Test coverage provided in system tests to ensure the policy is being transmitted first thing on start up of the application.

### Notes for the reviewer

Not updating the docs yet since the feature is not ready to be used (needs backend support).

### More information

- [Jira ticket RUN-1404](https://snyksec.atlassian.net/browse/RUN-1404)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "prepare": "npm run build",
     "build": "tsc",
     "dev": "tsc-watch --project tsconfig.json --onSuccess 'node --inspect .'",
-    "debug": "tsc-watch --project tsconfig.json --onSuccess 'node --inspect --debug-brk .'",
+    "debug": "tsc-watch --project tsconfig.json --onSuccess 'node --inspect-brk .'",
     "lint": "eslint \"src/**/*.ts\" && (cd test && eslint \"**/*.ts\")"
   },
   "author": "snyk.io",
--- a/src/common/policy.ts
+++ b/src/common/policy.ts
@@ -1,0 +1,31 @@
+import { existsSync, readFile } from 'fs';
+import { resolve as resolvePath } from 'path';
+import { promisify } from 'util';
+
+import { logger } from './logger';
+import { constructWorkloadAutoImportPolicy } from '../transmitter/payload';
+import { sendWorkloadAutoImportPolicy } from '../transmitter';
+import { config } from './config';
+
+const readFileAsync = promisify(readFile);
+
+export async function loadAndSendWorkloadAutoImportPolicy(): Promise<void> {
+  try {
+    /** This path is set in snyk-monitor during installation/deployment and is defined in the Helm chart. */
+    const userProvidedRegoPolicyPath = resolvePath(
+      config.IMAGE_STORAGE_ROOT,
+      'policies',
+      'workload-auto-import.rego',
+    );
+    if (!existsSync(userProvidedRegoPolicyPath)) {
+      logger.info({}, 'Rego policy file does not exist, skipping loading');
+      return;
+    }
+
+    const regoPolicy = await readFileAsync(userProvidedRegoPolicyPath, 'utf8');
+    const payload = constructWorkloadAutoImportPolicy(regoPolicy);
+    await sendWorkloadAutoImportPolicy(payload);
+  } catch (err) {
+    logger.error({ err }, 'Unexpected error occurred while loading workload auto-import policy');
+  }
+}
--- a/src/transmitter/index.ts
+++ b/src/transmitter/index.ts
@@ -10,6 +10,7 @@ import {
   IRequestError,
   ScanResultsPayload,
   IDependencyGraphPayload,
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "start": "bin/start",
     "prepare": "npm run build",
     "build": "tsc",
-    "dev": "tsc-watch --project tsconfig.json --onSuccess 'node --inspect .'",
+    "dev": "tsc-watch --project tsconfig.json --onSuccess 'node --inspect-brk .'",
```
dkontorovskyy, line 22: `npm run dev` should be for running project, not waiting for brk point. 
`npm run debug` makes for sense for that 
IMHO

src/common/policy.ts
```diff
@@ -0,0 +1,36 @@
+import { existsSync } from 'fs';
```
dkontorovskyy, line 1: Double fs import can be shortend on lines 1 and 3 :)

src/common/policy.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export async function loadAndSendWorkloadAutoImportPolicy(): Promise<void> {
+  try {
+    /** This path is set in snyk-monitor during installation/deployment and is defined in the Helm chart. */
+    const userProvidedRegoPolicyPath = resolvePath(
+      config.IMAGE_STORAGE_ROOT,
+      'policies',
+      'workload-auto-import.rego',
+    );
+    if (!existsSync(userProvidedRegoPolicyPath)) {
```
dkontorovskyy, line 21: you can also use async exists for non-blocking purposes?

src/common/policy.ts
```diff
@@ -0,0 +1,36 @@
...
+      logger.info({}, 'Rego policy file does not exist, skipping loading');
+      return;
+    }
+
+    const regoPolicy = await readFileAsync(userProvidedRegoPolicyPath, 'utf8');
+    const payload = constructWorkloadAutoImportPolicy(regoPolicy);
+    try {
```
dkontorovskyy, line 28: I would recommend not doing nested `try` as it's harder to handle errors and also taking a lot of resources from process

src/transmitter/index.ts
```diff
@@ -69,6 +70,30 @@ export async function sendWorkloadMetadata(payload: IWorkloadMetadataPayload): P
...
+  try {
+    logger.info(
+      { userLocator: payload.userLocator, cluster: payload.cluster, agentId: payload.agentId },
+      'attempting to send workload auto-import policy',
+    );
+
+    const { response, attempt } = await retryRequest('post', `${upstreamUrl}/api/v1/policy`, payload);
+    if (!isSuccessStatusCode(response.statusCode)) {
+      throw new Error(`${response.statusCode} ${response.statusMessage}`);
+    } else {
```
dkontorovskyy, line 83: no need in `else` it won't continue this path after throwing error

package.json
ivanstanev: Committed by accident while testing locally, but indeed useful as `debug`

src/common/policy.ts
ivanstanev: Good spot

src/transmitter/index.ts
```diff
@@ -69,6 +70,30 @@ export async function sendWorkloadMetadata(payload: IWorkloadMetadataPayload): P
...
 }
 
+export async function sendWorkloadAutoImportPolicy(payload: WorkloadAutoImportPolicyPayload): Promise<void> {
+  try {
+    logger.info(
+      { userLocator: payload.userLocator, cluster: payload.cluster, agentId: payload.agentId },
+      'attempting to send workload auto-import policy',
+    );
+
+    const { response, attempt } = await retryRequest('post', `${upstreamUrl}/api/v1/policy`, payload);
```
jmickey, line 80: Do we want to wait for #664 to be merged before we merge this one, and we can use the queue? Or do we want to skip the queue for this payload?

dkontorovskyy: Another things, does homebase has `/api/v1/policy` route? Or this PR is actually blocked until we add `/api/v1/policy` into homebase?

src/common/policy.ts
jmickey: I though async exists was deprecated?

src/transmitter/index.ts
ivanstanev: Doesn't exist yet but doesn't have to be blocking since it will print an error that it can't send the policy only once on deployment (assuming we can tolerate that until the Homebase endpoint is added)

dkontorovskyy (approved):

PR #718, opened 2021-04-19
Feat/deployment config

ivanstanev:

### What this does

Add support for [DeploymentConfig](https://docs.openshift.com/container-platform/4.7/applications/deployments/what-deployments-are.html) on OpenShift.

Before we set up Informers in the cluster for all the different workloads (e.g. Deployment, ReplicaSet), specifically for DeploymentConfig we now do a small query to the API server to understand if it's supported. If the API request returns successfully (non-404 code) then DeploymentConfig is supported and we set up a Watch following the routes in the [API reference](https://docs.openshift.com/container-platform/4.7/rest_api/workloads_apis/deploymentconfig-apps-openshift-io-v1.html). If the K8s API returns 404 (meaning the resource is unknown) then we skip creating a DeploymentConfig Informer.

What was added:
- an integration test for OpenShift 4 that applies a DeploymentConfig and checks that the workload was scanned
- TypeScript types for `DeploymentConfigV1` based on the [API reference](https://docs.openshift.com/container-platform/4.7/rest_api/workloads_apis/deploymentconfig-apps-openshift-io-v1.html)
- some functions are now `async` to allow querying the API server for supported workloads before deciding whether to set an Informer; we also await `informer.start()` calls since it returns a Promise

### More information

Commit by commit 🙏 

- [Jira ticket RUN-1505](https://snyksec.atlassian.net/browse/RUN-1505)

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "lint": "eslint \"src/**/*.ts\" && (cd test && eslint \"**/*.ts\")"
   },
   "dependencies": {
-    "@kubernetes/client-node": "^0.14.2",
+    "@kubernetes/client-node": "^0.14.3",
     "@snyk/dep-graph": "^1.28.0",
     "async": "^3.2.0",
     "aws-sdk": "^2.870.0",
--- a/src/supervisor/types.ts
+++ b/src/supervisor/types.ts
@@ -1,6 +1,15 @@
-import { IncomingMessage }  from 'http';
-import { AppsV1Api, BatchV1Api, BatchV1beta1Api, CoreV1Api, KubeConfig,
-  V1ObjectMeta, V1OwnerReference, V1PodSpec } from '@kubernetes/client-node';
+import { IncomingMessage } from 'http';
+import {
+  AppsV1Api,
+  BatchV1Api,
+  BatchV1beta1Api,
+  CoreV1Api,
+  CustomObjectsApi,
+  KubeConfig,
+  V1ObjectMeta,
+  V1OwnerReference,
+  V1PodSpec,
+} from '@kubernetes/client-node';
 
 export enum WorkloadKind {
   Deployment = 'Deployment',
@@ -11,6 +20,7 @@ export enum WorkloadKind {
   CronJob = 'CronJob',
   ReplicationController = 'ReplicationController',
   Pod = 'Pod',
+  DeploymentConfig = 'DeploymentConfig',
 }
 
 export interface IRequestError {
@@ -32,6 +42,7 @@ export interface IK8sClients {
   readonly coreClient: CoreV1Api;
   readonly batchClient: BatchV1Api;
   readonly batchUnstableClient: BatchV1beta1Api;
+  readonly customObjectsClient: CustomObjectsApi;
 }
 
 export class K8sClients implements IK8sClients {
@@ -41,12 +52,17 @@ export class K8sClients implements IK8sClients {
   // TODO: Keep an eye on this! We need v1beta1 API for CronJobs.
   // https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning
   // CronJobs will appear in v2 API, but for now there' only v2alpha1, so it's a bad idea to use it.
+  // TODO: https://kubernetes.io/blog/2021/04/09/kubernetes-release-1.21-cronjob-ga/
+  // CronJobs are now GA in Kubernetes 1.21 in the batch/v1 API, we should add support for it!
   public readonly batchUnstableClient: BatchV1beta1Api;
+  /** This client is used to access Custom Resources in the cluster, e.g. DeploymentConfig on OpenShift. */
+  public readonly customObjectsClient: CustomObjectsApi;
 
   constructor(config: KubeConfig) {
     this.appsClient = config.makeApiClient(AppsV1Api);
     this.coreClient = config.makeApiClient(CoreV1Api);
     this.batchClient = config.makeApiClient(BatchV1Api);
     this.batchUnstableClient = config.makeApiClient(BatchV1beta1Api);
+    this.customObjectsClient = config.makeApiClient(CustomObjectsApi);
   }
 }
--- a/src/supervisor/watchers/handlers/deployment-config.ts
+++ b/src/supervisor/watchers/handlers/deployment-config.ts
@@ -1,0 +1,32 @@
+import { deleteWorkload } from './workload';
+import { WorkloadKind } from '../../types';
+import { FALSY_WORKLOAD_NAME_MARKER, V1DeploymentConfig } from './types';
+
+export async function deploymentConfigWatchHandler(
+  deploymentConfig: V1DeploymentConfig,
+): Promise<void> {
+  if (
+    !deploymentConfig.metadata ||
+    !deploymentConfig.spec ||
+    !deploymentConfig.spec.template.metadata ||
+    !deploymentConfig.spec.template.spec ||
+    !deploymentConfig.status
+  ) {
+    return;
+  }
+
+  const workloadName =
+    deploymentConfig.metadata.name || FALSY_WORKLOAD_NAME_MARKER;
+
+  await deleteWorkload(
+    {
+      kind: WorkloadKind.DeploymentConfig,
+      objectMeta: deploymentConfig.metadata,
+      specMeta: deploymentConfig.spec.template.metadata,
+      ownerRefs: deploymentConfig.metadata.ownerReferences,
+      revision: deploymentConfig.status.observedGeneration,
+      podSpec: deploymentConfig.spec.template.spec,
+    },
+    workloadName,
+  );
+}
--- a/src/supervisor/watchers/handlers/index.ts
+++ b/src/supervisor/watchers/handlers/index.ts
@@ -1,4 +1,5 @@
 import { makeInformer, ADD, DELETE, ERROR, UPDATE, KubernetesObject } from '@kubernetes/client-node';
+
 import { logger } from '../../../common/logger';
 import { WorkloadKind } from '../../types';
 import { podWatchHandler, podDeletedHandler } from './pod';
@@ -9,6 +10,7 @@ import { jobWatchHandler } from './job';
 import { replicaSetWatchHandler } from './replica-set';
 import { replicationControllerWatchHandler } from './replication-controller';
 import { statefulSetWatchHandler } from './stateful-set';
+import { deploymentConfigWatchHandler } from './deployment-config';
 import { k8sApi, kubeConfig } from '../../cluster';
 import * as kubernetesApiWrappers from '../../kuberenetes-api-wrappers';
 import { IWorkloadWatchMetadata, FALSY_WORKLOAD_NAME_MARKER } from './types';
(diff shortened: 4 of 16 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "lint": "eslint \"src/**/*.ts\" && (cd test && eslint \"**/*.ts\")"
   },
   "dependencies": {
-    "@kubernetes/client-node": "^0.14.2",
+    "@kubernetes/client-node": "^0.14.3",
```
dkontorovskyy, line 36: 0.14.3 removed the use of net-keepalive, so we also can remove dependencies we install in Dockefile
https://github.com/kubernetes-client/javascript/commit/a3e6c0aff670d2441c9b31691f36e3d159ad52ec

src/supervisor/types.ts
```diff
@@ -41,12 +52,17 @@ export class K8sClients implements IK8sClients {
   // TODO: Keep an eye on this! We need v1beta1 API for CronJobs.
   // https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning
   // CronJobs will appear in v2 API, but for now there' only v2alpha1, so it's a bad idea to use it.
+  // TODO: https://kubernetes.io/blog/2021/04/09/kubernetes-release-1.21-cronjob-ga/
```
dkontorovskyy, line 55: Can you reflect this with a ticket in Jira, please? :)

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,67 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
...
+      fieldSelector,
+      labelSelector,
+      limit,
+    );
+    return (
+      attemptedApiCall !== undefined &&
+      attemptedApiCall.response !== undefined &&
+      attemptedApiCall.response.statusCode !== undefined &&
+      attemptedApiCall.response.statusCode > 200 &&
+      attemptedApiCall.response.statusCode < 300
```
dkontorovskyy, line 136:
```suggestion
      Boolean(attemptedApiCall?.response?.statusCode > 200 && attemptedApiCall?.response?.statusCode > 200)
```

ivanstanev: `Object is possibly 'undefined'.ts(2532)` ;)

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,71 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
...
+  try {
+    const pretty = undefined;
+    const continueToken = undefined;
+    const fieldSelector = undefined;
+    const labelSelector = undefined;
+    const limit = 1; // Try to grab only a single object
+    const resourceVersion = undefined; // List anything in the cluster
+    const timeoutSeconds = 10; // Don't block the snyk-monitor indefinitely
+    const attemptedApiCall = await k8sApi.customObjectsClient.listNamespacedCustomObject(
+      'apps.openshift.io',
```
dkontorovskyy, line 123: nit: Why don't store this in an object and just reference GVK parts from the static map?

src/supervisor/watchers/handlers/index.ts
dkontorovskyy: oh, missed the first statement

src/supervisor/watchers/handlers/index.ts
```diff
@@ -113,11 +180,11 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
   informer.on(ERROR, (err) => {
     // Types from client library insists that callback is of type KubernetesObject
     if ((err as any).code === ECONNRESET_ERROR_CODE) {
-      logger.debug(`informer ${ECONNRESET_ERROR_CODE} occurred, restarting informer`);
+      logger.debug({}, `informer ${ECONNRESET_ERROR_CODE} occurred, restarting informer`);
```
dkontorovskyy, line 183: ❤️

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,71 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
     },
     listFactory: (namespace) => () => k8sApi.appsClient.listNamespacedStatefulSet(namespace),
   },
+  [WorkloadKind.DeploymentConfig]: {
+    /** https://docs.openshift.com/container-platform/4.7/rest_api/workloads_apis/deploymentconfig-apps-openshift-io-v1.html */
+    endpoint: '/apis/apps.openshift.io/v1/watch/namespaces/{namespace}/deploymentconfigs',
+    handlers: {
+      [DELETE]: deploymentConfigWatchHandler,
+    },
+    listFactory: (namespace) => () => k8sApi.customObjectsClient.listNamespacedCustomObject("apps.openshift.io", "v1", namespace, "deploymentconfigs"),
```
dkontorovskyy, line 102: nit: looks like we use single quotes everywhere

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,71 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
...
+
+  try {
+    const pretty = undefined;
+    const continueToken = undefined;
+    const fieldSelector = undefined;
+    const labelSelector = undefined;
+    const limit = 1; // Try to grab only a single object
+    const resourceVersion = undefined; // List anything in the cluster
+    const timeoutSeconds = 10; // Don't block the snyk-monitor indefinitely
+    const attemptedApiCall = await k8sApi.customObjectsClient.listNamespacedCustomObject(
```
minsiyang, line 122: I assume here is calling to the k8s client API to add the deploymentConfig functionality to the cluster? 😄

src/supervisor/watchers/handlers/index.ts
```diff
@@ -135,5 +202,5 @@ export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
     });
   }
 
-  informer.start();
+  await informer.start();
```
dkontorovskyy, line 205: ❤️

src/supervisor/watchers/handlers/types.ts
```diff
@@ -1,3 +1,5 @@
+import { V1ObjectMeta, V1PodTemplateSpec } from "@kubernetes/client-node";
```
dkontorovskyy, line 1: nit: use single quotes, please :)

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,71 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
...
+    listFactory: (namespace) => () => k8sApi.customObjectsClient.listNamespacedCustomObject("apps.openshift.io", "v1", namespace, "deploymentconfigs"),
+  },
 };
 
-export function setupInformer(namespace: string, workloadKind: WorkloadKind) {
+async function isSupportedWorkload(
+  namespace: string,
+  workloadKind: WorkloadKind,
+): Promise<boolean> {
+  if (workloadKind !== WorkloadKind.DeploymentConfig) {
```
dkontorovskyy, line 110: I'm just thinking if there is a way we can separate vanilla workloads from custom objects to avoid doing this if and not calling this function with vanilla workloads. Nothing comes to my mind for now

ivanstanev: We could probably re-use this function once we add support for the new API batch/v1 CronJobs - we might want to see if they're supported before adding an Informer

src/supervisor/watchers/handlers/types.ts
ivanstanev: `prettier` better!

src/supervisor/watchers/handlers/index.ts
ivanstanev: This queries the API to see if DeploymentConfigs can be returned by Kubernetes. If the API returns something, then we know they're supported and we can later set up an Informer to watch for changes. If this query fails then DeploymentConfig is likely unsupported

src/supervisor/watchers/handlers/index.ts
ivanstanev: Could you share an example code? Not sure what you mean

src/supervisor/watchers/handlers/index.ts
dkontorovskyy: sweeeeeet :)

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,71 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
...
+    logger.info(
+      { error, workloadKind },
+      'Failed on Kubernetes API call to list DeploymentConfig',
+    );
+    return false;
+  }
+}
+
+export async function setupInformer(namespace: string, workloadKind: WorkloadKind): Promise<void> {
+  const isSupported = await isSupportedWorkload(namespace, workloadKind);
```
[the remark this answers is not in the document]
ivanstanev: 💡 
But it would complicate the code a bit, we now need to check this only once and we need to differentiate between snyk-monitor scanning the cluster or snyk-monitor scanning a single namespace (won't have permission to query the cluster if scanning a single namespace).

src/supervisor/watchers/handlers/index.ts
```diff
@@ -91,9 +93,71 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
...
+      labelSelector,
+      limit,
+      resourceVersion,
+      timeoutSeconds,
+    );
+    return (
+      attemptedApiCall !== undefined &&
+      attemptedApiCall.response !== undefined &&
+      attemptedApiCall.response.statusCode !== undefined &&
+      attemptedApiCall.response.statusCode > 200 &&
```
kat1906, line 139: is `200` not valid here?

src/supervisor/watchers/handlers/types.ts
kat1906: can we set up a linting rule for this?

src/supervisor/watchers/handlers/index.ts
ivanstanev: Nice catch!

src/supervisor/watchers/handlers/index.ts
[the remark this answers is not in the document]
ivanstanev: Opting not to do it because it would complicate the code quite a bit, LMK if you think it's a really bad idea and the benefits would outweigh the downsides 😉

src/supervisor/watchers/handlers/deployment-config.ts
```diff
@@ -0,0 +1,32 @@
...
+    !deploymentConfig.spec.template.spec ||
+    !deploymentConfig.status
+  ) {
+    return;
+  }
+
+  const workloadName =
+    deploymentConfig.metadata.name || FALSY_WORKLOAD_NAME_MARKER;
+
+  await deleteWorkload(
```
kat1906, line 21: how come the workload is deleted?

kat1906: oh it only has a delete command!

kat1906: should `delete` be in the name of the handler? (I see that this is not what we have done in the others apart from pod but maybe it should be for consistency?)

kat1906 (approved):

src/supervisor/watchers/handlers/deployment-config.ts
ivanstanev: @kat1906 Yes I think it's the second time the names trip us up here, I'll open a follow up PR to improve them 🙏 The `deleteWorkload()` here only sends an event to Upstream that this workload was deleted, it doesn't actually do anything to the workload in K8s 😄

PR #730, opened 2021-04-26
add support for pull oci images

chen-keinan:

### What this does
Added fallback logic for pulling oci-archive image incase docker-archive image pull fails after x retries.

### DIFF
--- a/src/scanner/images/index.ts
+++ b/src/scanner/images/index.ts
@@ -4,7 +4,7 @@ import { DepGraph, legacy } from '@snyk/dep-graph';
 
 import { logger } from '../../common/logger';
 import { pull as skopeoCopy, getDestinationForImage } from './skopeo';
-import { IPullableImage, IScanImage } from './types';
+import {IPullableImage, IScanImage, SkopeoRepositoryType} from './types';
 import { IScanResult } from '../types';
 import {
   buildDockerPropertiesOnDepTree,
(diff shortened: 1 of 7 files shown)


oci_changes.patch
```diff
@@ -0,0 +1,222 @@
+diff --git a/.gitignore b/.gitignore
```
ivanstanev, line 1: This file might be checked in by accident

chen-keinan: yes removing it

ivanstanev (approved):

src/scanner/images/index.ts
```diff
@@ -13,30 +13,43 @@ import {
...
+        } catch (error) {
+            logger.error({error, image: image.imageWithDigest?? image.imageName}, 'failed to pull image docker/oci archive image');
+        }
+    }
+    return pulledImages;
 }
 
-export function getImagesWithFileSystemPath(images: IScanImage[]): IPullableImage[] {
-  return images.map((image) => ({ ...image, fileSystemPath: getDestinationForImage(image.imageName) }));
+export function getImagesWithFileSystemPath(images: IScanImage[]): { imageName: string; skopeoRepoType: SkopeoRepositoryType; fileSystemPath: string; imageWithDigest?: string }[] {
```
dkontorovskyy, line 51: I would suggest to create a type for this object
```js
{ imageName: string; skopeoRepoType: SkopeoRepositoryType; fileSystemPath: string; imageWithDigest?: string }
```
and use it here for readability purposes 🤗

src/scanner/images/index.ts
```diff
@@ -75,10 +88,11 @@ export function getImageParts(imageWithTag: string) : {imageName: string, imageT
 export async function scanImages(images: IPullableImage[]): Promise<IScanResult[]> {
   const scannedImages: IScanResult[] = [];
 
-  for (const { imageName, fileSystemPath, imageWithDigest } of images) {
+  for (const { imageName, fileSystemPath, imageWithDigest, skopeoRepoType } of images) {
     try {
       const shouldIncludeAppVulns = true;
-      const dockerArchivePath = `docker-archive:${fileSystemPath}`;
+       const archiveType= skopeoRepoType == SkopeoRepositoryType.DockerArchive?"docker-archive":"oci-archive";
```
dkontorovskyy, line 94: I'm curious how this passed our linter 🤔  We definitely need to fix our linter and maybe introduce `prettier` in a separate PR. 
This line will require only adding some spaces between `=`, `?` and `:`.
@chen-keinan it's great that you were able to help to catch this :)

ivanstanev: It doesn't need spaces but I didn't comment on style in this PR because it's pointless, we can fix all of this by introducing `prettier` to format the code rather than asking people to remember

PR #733, opened 2021-04-27
chore: introduce code formatter and style

chen-keinan:

### What this does
Introduce code formatter and style


.circleci/config.yml
```diff
@@ -767,6 +767,44 @@ jobs:
                 name: Notify Slack on failure
                 when: on_fail
         working_directory: ~/kubernetes-monitor
+    lint:
```
ivanstanev, line 770: We use the `circleci` binary to pack the configuration into this file. The CircleCI config is split into several files and then combined into this one as noted in the README: https://github.com/snyk/kubernetes-monitor/tree/staging/.circleci Could you also ensure the changes are added outside of this file and then the config.yml is generated?

.circleci/config.yml
```diff
@@ -985,6 +1022,18 @@ workflows:
                         ignore:
                             - staging
                             - master
+            - lint:
```
ivanstanev, line 1025: Should we run both linting and code_formatting on MERGE_TO_STAGING too?

package-lock.json
```diff
@@ -1,7 +1,14120 @@
 {
   "name": "@snyk/kubernetes-monitor",
+  "lockfileVersion": 2,
```
ivanstanev, line 3: I think you're running `npm` version 7 which most of us don't use yet, could you ensure we're still running version 6? I think most of our repos are not yet ready for version 7, even though it looks like `snyk-monitor` passed the tests so it will likely be OK!

ivanstanev (commented):
Great change 👏

.circleci/config.yml
chen-keinan: sure why not , should be align with unit test

ivanstanev (approved):

PR #742, opened 2021-05-01
fix: fixes #739 set securityContext.fsGroup at pod level

### LINKED ISSUE #739: snyk monitor unable to pull images when using PVC due to permission issues
This might be related to #697 but using PVC (created by snyk monitor helm chart), I came across issue where snyk monitor was not able to write to` /var/tmp`. Not sure why but even with permissions set by init container it won't allow a a container running as non root to write to it. I think PVC mount perms work differently.

So I solved this is by setting fsgroup on pod level which worked i.e.

```
securityContext:
        fsGroup: 2000
```

And I also removed init container as with this context we don't need it anymore. I don't understand what was the need for this container in the first place if we could use `fsgroup`
I have not tested it with non-pvc volume but I think it should work.

jmickey:
Instead of using an `initContainer` we instad set the `fsGroup` within the pod `securityContext`. Kubernetes will **create** this supplimentary group and assign it to all processes running within the pod.

Fixes #739

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -18,23 +18,10 @@ spec:
         app.kubernetes.io/name: {{ include "snyk-monitor.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      securityContext:
+        fsGroup: 2000
       serviceAccountName: {{ include "snyk-monitor.name" . }}
       restartPolicy: Always
-      initContainers:
-        - name: volume-permissions
-          image: "{{ .Values.initContainerImage.repository }}:{{ .Values.initContainerImage.tag }}"
-          command : ['sh', '-c', 'chmod -R go+rwX /var/tmp || true']
-          volumeMounts:
-            - name: temporary-storage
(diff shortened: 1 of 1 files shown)


snyk-monitor/templates/deployment.yaml
```diff
@@ -18,23 +18,10 @@ spec:
         app.kubernetes.io/name: {{ include "snyk-monitor.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      securityContext:
+        fsGroup: 2000
```
ivanstanev, line 22: Unfortunately I don't think this works on OpenShift, the group IDs are randomly allocated per namespace and they're some really high numbers (in the millions). We should have a look into OS's documentation regarding these 🤔

jmickey: The group doesn't have to exist for this to work. Kubernetes actually creates the `fsGroup` when the pod starts and assigns it as a supplimentary group to all running processes. The groupIds that OS generates won't matter.

rumi-spock (approved):
This works on our eks deployment.

ivanstanev (approved):

PR #763, opened 2021-05-18
[RUN-1515] Tidy up docker image

kat1906:
Exclude files which are not needed from Docker by adding to .dockerignore.

### What this does

Tidies up the Docker image by excluding files which are not needed by adding to .dockerignore.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,50 +1,57 @@
-.github
-.nyc_output
-.dockerignore
 .git*
-Dockerfile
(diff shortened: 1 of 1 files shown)


jmickey (approved):

ivanstanev (changes requested):
Could we make this a `feat` or `fix` so that it can release a new image with these changes?

ivanstanev (approved):

PR #769, opened 2021-05-24
feat: re-use sockets when sending data to kubernetes upstream

dkontorovskyy:

### What this does
Try to use `keepAlive` during `kubernetes-upstream` requests.

This PR avoids using `keepAlive` usage for instances that are using proxies.

### DIFF
--- a/src/transmitter/index.ts
+++ b/src/transmitter/index.ts
@@ -1,7 +1,11 @@
+import { parse } from 'url';
 import { queue } from 'async';
 import * as needle from 'needle';
-import { NeedleResponse, NeedleHttpVerbs, NeedleOptions } from 'needle';
 import * as sleep from 'sleep-promise';
+import { Agent as HttpAgent } from 'http';
+import { Agent as HttpsAgent } from 'https';
+import { NeedleResponse, NeedleHttpVerbs, NeedleOptions } from 'needle';
+
 import { logger } from '../common/logger';
 import { config } from '../common/config';
 import {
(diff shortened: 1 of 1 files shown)


src/transmitter/index.ts
```diff
@@ -227,6 +244,7 @@ async function retryRequest(
   const options: NeedleOptions = {
     json: true,
     compressed: true,
+    agent,
   };
   if (config.HTTP_PROXY || config.HTTPS_PROXY) {
     options.agent = getProxyAgent(config, url);
```
ivanstanev, line 250: Will also need to ensure a keep-alive agent is used in the response to `getProxyAgent()` 🙏

src/transmitter/index.ts
```diff
@@ -28,6 +31,20 @@ interface HomebaseRequest {
 const upstreamUrl =
   config.INTEGRATION_API || config.DEFAULT_KUBERNETES_UPSTREAM_URL;
 
+let agent = new HttpAgent({
+  keepAlive: true,
+  // Infinity is read as 50 sockets
```
ivanstanev, line 36: Are you sure this is the case? I couldn't find docs that say Infinity === 50, it looks like it really means unlimited sockets 🤔

src/transmitter/index.ts
dkontorovskyy: As discussed, not doing this for proxy connections yet

ivanstanev (approved):

PR #777, opened 2021-05-27
feat: Helm flag to set the Orgs to use for workload auto import/delete

ivanstanev:

### What this does

The workload auto-import and auto-delete policy now comes as a preset. In order to use it, users should provide a list of Snyk Organization public IDs so that snyk-monitor knows which Orgs to use for importing or deleting scanned images.

### DIFF
--- a/snyk-monitor/templates/configmap.yaml
+++ b/snyk-monitor/templates/configmap.yaml
@@ -3,9 +3,35 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ .Release.Name }}-excluded-namespaces
+  labels:
+    app.kubernetes.io/name: {{ include "snyk-monitor.name" . }}
+    helm.sh/chart: {{ include "snyk-monitor.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
   excludedNamespaces: |-
     {{- range .Values.excludedNamespaces }}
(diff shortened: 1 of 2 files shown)


snyk-monitor/templates/configmap.yaml
```diff
@@ -3,9 +3,35 @@ apiVersion: v1
...
+    helm.sh/chart: {{ include "snyk-monitor.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
   excludedNamespaces: |-
     {{- range .Values.excludedNamespaces }}
     {{ . }}
     {{- end }}
 {{ end }}
+---
```
dkontorovskyy, line 17: Do we need `{{ if .Values.policyOrgs }}`? Or we want this policy be set for every kubernete-monitor by default?

dkontorovskyy (commented):
Nice change 👏 

Do we also need to update README.md with new Helm values?

ivanstanev:
> Nice change 👏
> 
> Do we also need to update README.md with new Helm values?

There isn't anything in the README yet because we haven't gotten to the task of adding the docs yet - that's a pre-GA task 🙂

snyk-monitor/templates/configmap.yaml
ivanstanev: We want it set by default to make it easier for on-boarding

kat1906 (approved):

PR #787, opened 2021-06-01
docs: add info on --reuse-values to README

kat1906:

### What this does

Add info on `--reuse-values` option to README so users know they do not need to keep setting the values on every upgrade.

### DIFF
--- a/snyk-monitor/README.md
+++ b/snyk-monitor/README.md
@@ -188,11 +188,18 @@ By default, `snyk-monitor` does not scan containers that are internal to Kuberne
 * `kube-system`
 * `local-path-storage`
 
-If you prefer to override this, you can add your own list of namespaces to exclude by setting the `excludedNamespaces` to your own list.
-For example:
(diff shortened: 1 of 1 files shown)


snyk-monitor/README.md
```diff
@@ -188,12 +188,19 @@ By default, `snyk-monitor` does not scan containers that are internal to Kuberne
...
+If you prefer to override this, you can add your own list of namespaces to exclude by setting the `excludedNamespaces` to your own list. For example:
 ```yaml
 --set excludedNamespaces={kube-node-lease,kube-public,local-path-storage,some_namespace}
 ```
 
+## Upgrading
+
+When upgrading the Snyk monitor, if you would like to reuse the last release's values and merge in any overrides from the command line via --set and -f, you can use the option `--reuse-values`. For example:
+```bash
+helm upgrade snyk-monitor snyk-charts/snyk-monitor --reset-values
```
ivanstanev, line 200: Just to double check, is this example supposed to say `--reuse-values`?

kat1906: 🤦

ivanstanev (approved):

Issue #809, opened 2021-06-28
[🐛] snyk-monitor Chart: Specify custom Configmap with rego policy for import/export


richerve:
- `kubernetes-monitor version: v1.64.1` 
- `Cloud runtime: custom kops based`

### Expected behaviour

The helm chart `workloadPoliciesMap` value will allow to specify a configmap including custom rego policies under the `workload-events.rego` key. Changing that value in `values.yaml` must not have any effect in the included default policy.

### Actual behaviour

Here's a couple of options I have in mind:

1. (preferred) Use `workloadPoliciesMap` in values.yaml to point to a pre-created custom ConfigMap and mount that in the deployment. The ConfigMap in the Chart containing the default policy, should have a fixed name, not configurable (but documented, of course).
2. Add a new option in values.yaml to point to the external ConfigMap and adjust the Deployment accordingly.

### Steps to reproduce

Set `workloadPoliciesMap` to a name of an existing ConfigMap having the custom policies. The helm release will fail as a ConfigMap with the same name is already part of the helm release.

kat1906:
Hi @richerve thanks for raising this issue.

I'll take it to the team and we'll get back to you 🙂

richerve:
I'm working in a PR already that I'm planning to submit, if that's ok

kat1906:
Even better! Thanks

richerve:
@kat1906 https://github.com/snyk/kubernetes-monitor/pull/812

dkontorovskyy:
@richerve it's merged :) Can you go and check `v1.64.4`? :)

dkontorovskyy:
@richerve have you had any chance to check? 🤗

richerve:
> @richerve have you had any chance to check? hugs

yes, all looks good, thanks

dkontorovskyy:
Closing it as fixed :) 

Reached you out via email to send you swag for the amazing PR you've opened!

PR #812, opened 2021-07-01
snyk-monitor: Add defaultWorkloadPolicies option

richerve:

### What this does

To provide an option to pass an external ConfigMap containing workload policies. Currently it's not possible to provide a custom ConfigMap, as the default one is created and mounted by default. More info in https://github.com/snyk/kubernetes-monitor/issues/809

### Notes for the reviewer

I decided introduce a slight change to the values.yaml interface regarding workload policies, but in terms of default behaviour, those changes don't change anything. The default policy will still be installed by default as it happens in the current version.

`workloadPoliciesMap` changes behaviour to represent an external ConfigMap name instead of the default map name.

I'm of course open for changes on the interface, but in my opinion this is the most flexible setup.

This change might affect documentation on: https://support.snyk.io/hc/en-us/articles/360020835037-Automatic-import-deletion-of-Kubernetes-workloads-projects, although with the current version, the last section of that documentation where an external ConfigMap is created, doesn't really work.

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -166,7 +166,11 @@ spec:
             optional: true
         - name: workload-policies
           configMap:
+            {{- if .Values.workloadPoliciesMap }}
             name: {{ .Values.workloadPoliciesMap }}
+            {{- else }}
+            name: {{ .Values.defaultWorkloadPoliciesMap }}
+            {{- end }}
             optional: true
         - name: registries-conf
           configMap:
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -8,10 +8,17 @@ monitorSecrets: snyk-monitor
 certsConfigMap: snyk-monitor-certs
 registriesConfConfigMap: snyk-monitor-registries-conf
 
-# The ConfigMap to use for loading policies into snyk-monitor.
-workloadPoliciesMap: snyk-monitor-workload-policies
+# An external ConfigMap to use for loading policies into snyk-monitor.
+# If not set, defaultWorkloadPoliciesMap will be used.
+workloadPoliciesMap: ""
+
 # A list of Snyk Organization public IDs to let snyk-monitor know in which Organization to auto-import and auto-delete scanned images.
-policyOrgs:
+# This is used to populate the default workload policy file.
+policyOrgs: []
+
+# Default rego workload policies to install.
+# Setting `workloadPoliciesMap` will overwrite this.
+defaultWorkloadPoliciesMap: snyk-monitor-workload-policies
 
 # One of: Cluster, Namespaced
 # Cluster - creates a ClusterRole and ClusterRoleBinding with the ServiceAccount
--- a/snyk-monitor/templates/configmap.yaml
+++ b/snyk-monitor/templates/configmap.yaml
@@ -14,11 +14,12 @@ data:
     {{ . }}
     {{- end }}
 {{ end }}
+{{ if not .Values.workloadPoliciesMap }}
 ---
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ .Values.workloadPoliciesMap }}
+  name: {{ .Values.defaultWorkloadPoliciesMap }}
   labels:
     app.kubernetes.io/name: {{ include "snyk-monitor.name" . }}
     helm.sh/chart: {{ include "snyk-monitor.chart" . }}
@@ -35,3 +36,4 @@ data:
     workload_events {
       input.kind != "Job"
     }
+{{ end }}


dkontorovskyy:
@richerve thanks much for your contribution 🎊 
The team will be reviewing and testing it shortly 🧪

snyk-monitor/values.yaml
```diff
@@ -8,10 +8,17 @@ monitorSecrets: snyk-monitor
...
+# When using this the default policy shouldn't be deployed, set `defaultWorkloadPolicy.enabled: false`.
+workloadPoliciesMap: ''
+
+# Default rego workload policies to install.
+# When setting `workloadPoliciesMap`, this should be disabled.
+defaultWorkloadPolicies:
+  enabled: true
+  name: snyk-monitor-workload-policies
+  # A list of Snyk Organization public IDs to let snyk-monitor know in which Organization to auto-import and auto-delete scanned images.
+  policyOrgs: []
```
snaftaly, line 21: Could you please also add support for policyOrgs to be defined in the global scope, so that this change will not break current functionality?
This will also require a change in the configmap, so that we have a fallback to the global policyOrgs in case the defaultWorkloadPolicies.policyOrgs is not defined.

snyk-monitor/templates/deployment.yaml
```diff
@@ -166,7 +166,11 @@ spec:
             optional: true
         - name: workload-policies
           configMap:
+            {{- if .Values.defaultWorkloadPolicies.enabled }}
```
snaftaly, line 169: Would it be possible to flip the order of the if statement, so that instead having the following check `if .Values.defaultWorkloadPolicies.enabled`, check for the existence of workloadPoliciesMap (i.e. check if workloadPoliciesMap is set)?

This also applies to [line 18](https://github.com/snyk/kubernetes-monitor/pull/812/files#diff-037fa431fc058f8bafd63b4732e44bfeb414e6f93fc9dc3c57f5a0ff047c2c71R18) in configmap.yaml, where instead of checking for the enabled flag, we could possibly check if 'workloadPoliciesMap' is not set.

This way we can possibly remove the 'enabled' flag from defaultWorkloadPolicies.

snaftaly (commented):
@richerve Thank you very much for this fix 🙇 
We have tested it and it indeed solves the issue you reported.
We have reviewed the code and added a few comments. 

Also, since we are using semantic versioning for our releases, code you please add the following prefix in the commit title: "fix: <you commit title>"?

snyk-monitor/templates/deployment.yaml
richerve: ok, I get what you're saying, will try that

snyk-monitor/values.yaml
richerve: Is the idea of not changing the current api? (I consider values.yaml as an api). If we are keeping policyOrgs in global scope and the enabled key in default policies can be removed, the there's probably better to just set `defaultWorkloadPoliciesMap` to the name of the default policy and verify if WorkloadPoliciesMap is defined to override that. What do you think?

richerve:
> @richerve Thank you very much for this fix bow
> We have tested it and it indeed solves the issue you reported.
> We have reviewed the code and added a few comments.
> 
> Also, since we are using semantic versioning for our releases, code you please add the following prefix in the commit title: "fix: "?

@snaftaly do you mean renaming the PR (that eventually would be the merge commit)? or any specific commit?

snaftaly:
> > @richerve Thank you very much for this fix bow
> > We have tested it and it indeed solves the issue you reported.
> > We have reviewed the code and added a few comments.
> > Also, since we are using semantic versioning for our releases, code you please add the following prefix in the commit title: "fix: "?
> 
> @snaftaly do you mean renaming the PR (that eventually would be the merge commit)? or any specific commit?

@richerve we are using [semantic release](https://github.com/semantic-release/semantic-release), which means the commit message itself needs to have this prefix (you can amend the existing commit and force push it).

snyk-monitor/values.yaml
snaftaly: Sounds like a good idea 👍

richerve:
@snaftaly done. I removed the default workload policies map and instead it only has the name that will be referenced locally. The behaviour now is that if `workloadPoliciesMap` is defined, it will overwrite the default, as it was originally intended.

snyk-monitor/values.yaml
```diff
@@ -8,18 +8,25 @@ monitorSecrets: snyk-monitor
...
+defaultWorkloadPoliciesMap: snyk-monitor-workload-policies
 
 # One of: Cluster, Namespaced
 # Cluster - creates a ClusterRole and ClusterRoleBinding with the ServiceAccount
 # Namespaced - creates a Role and RoleBinding with the ServiceAccount
 scope: Cluster
 
 # The endpoint that being used to transmit monitored information
-integrationApi: ""
+integrationApi: ''
```
dkontorovskyy, line 29: @richerve a tiny question, what is the difference between `''` and `""` in Helm world?

richerve: Oh sorry, that was my linter working. This is just yaml, there's no real distinction between '' and "". In helm values.yaml is static, there's no interpolation whatsoever so the semantics of yaml remain. Do you want me to revert to `""`?

snyk-monitor/values.yaml
```diff
@@ -8,18 +8,25 @@ monitorSecrets: snyk-monitor
 certsConfigMap: snyk-monitor-certs
 registriesConfConfigMap: snyk-monitor-registries-conf
 
-# The ConfigMap to use for loading policies into snyk-monitor.
-workloadPoliciesMap: snyk-monitor-workload-policies
+# An external ConfigMap to use for loading policies into snyk-monitor.
+# When using this the default policy shouldn't be deployed, set `defaultWorkloadPolicy.enabled: false`.
```
snaftaly, line 12: Could you please update this doc line? 
I would suggest to write the following - "If not set, defaultWorkloadPoliciesMap will be used".

snaftaly (commented):
Recent change looks good 👍 
Could you please update the comment for workloadPoliciesMap?

snyk-monitor/values.yaml
richerve: sure, no problem

snyk-monitor/values.yaml
dkontorovskyy: @richerve I think like we've already asked you for a lot of changes 😄 
Just this change doesn't have anything to do with the initial issue and I would like to leave this change out if possible 🙏

snyk-monitor/values.yaml
richerve: done

dkontorovskyy:
Hey, @richerve we will merge your PR into an internal branch that will allow us to run tests and merge further on the pipeline. We will keep you updated. Thanks a lot for this amazing contribution.

PR #848, opened 2021-08-12
fix: stabilise the monitor agent id

jmickey:
Rather than retrying the request at most 3 times, and then simply setting the agentId to an automatically generated UUID, we instead keep retrying the request indefinitely until we're able to retrieve the Kubernetes Deployment ID.

The retry logic will backoff exponentially up to a maximum configured value in seconds.

### DIFF
--- a/src/supervisor/agent.ts
+++ b/src/supervisor/agent.ts
@@ -1,7 +1,7 @@
 import { config } from '../common/config';
 import { logger } from '../common/logger';
 import { k8sApi } from './cluster';
-import { retryKubernetesApiRequest } from './kuberenetes-api-wrappers';
+import { retryKubernetesApiRequestIndefinitely } from './kuberenetes-api-wrappers';
 
 export async function setSnykMonitorAgentId(): Promise<void> {
   const name = config.DEPLOYMENT_NAME;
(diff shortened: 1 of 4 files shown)


src/supervisor/kuberenetes-api-wrappers.ts
```diff
@@ -26,6 +27,43 @@ export async function retryKubernetesApiRequest<ResponseType>(
...
+ *
+ * @param func function to retry
+ * @param maxSleepDuration maximum sleep duration in seconds (e.g. 300)
+ * @returns Promise<ResponseType>
+ */
+export async function retryKubernetesApiRequestIndefinitely<ResponseType>(
+  func: IKubernetesApiFunction<ResponseType>,
+  maxSleepDuration: number,
+): Promise<ResponseType> {
+  var attempts: number = 1;
```
dkontorovskyy, line 43:
```suggestion
  let attempts: number = 1;
```

src/supervisor/agent.ts
```diff
@@ -20,8 +20,9 @@ async function getSnykMonitorDeploymentUid(
   namespace: string,
 ): Promise<string | undefined> {
   try {
-    const attemptedApiCall = await retryKubernetesApiRequest(() =>
-      k8sApi.appsClient.readNamespacedDeployment(name, namespace),
+    const attemptedApiCall = await retryKubernetesApiRequestIndefinitely(
+      () => k8sApi.appsClient.readNamespacedDeployment(name, namespace),
+      300,
```
dkontorovskyy, line 25: Should we consider fetching this number from config? 
Also, I would prefer this number to be assigned to a variable, so we would omit using magic numbers

dkontorovskyy:
Nice work! 

Couple more questions from me:
- Can you remove `package-lock.json` from the PR? It makes sense to include `package-lock.json` when there are changes in `package.json`
- Is there a way to have tests for this change?

jmickey:
@dkontorovskyy 

I have removed the `package-lock.json`.

Regarding testing - I'm not really sure, difficult one to test.

kat1906 (approved):

PR #871, opened 2021-09-15
AWS ServiceAccountToken not accessible.

tektite-io:
### What this does

AWS ServiceAccountToken is not readable by the snyk-monitor process.
Using fsGroup and projected volume to solve the issue.

See similar problem: https://github.com/kubernetes-sigs/external-dns/pull/1185

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -27,6 +27,10 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
+      {{- if ne .Values.securityContext.fsGroup "" }}
+      securityContext:
+        fsGroup: 65534
+      {{- end }}
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


snyk-monitor/templates/deployment.yaml
```diff
@@ -172,6 +176,13 @@ spec:
             name: {{ .Values.defaultWorkloadPoliciesMap }}
             {{- end }}
             optional: true
+        {{- if .Values.volumes.projected.serviceAccountToken }}
+        - name: token-vol
+          projected:
+            sources:
+            - serviceAccountToken:
+                path: token  
+        {{- end }}
```
ivanstanev, line 185: Hey @tektite-io I was wondering if this volume also needs to be mounted with a `volumeMount`? The example in the docs both declares the `volume` and a `volumeMount` - https://kubernetes.io/docs/concepts/storage/volumes/#example-configuration-secret-downwardapi-configmap
And if we have to mount this volume, where should it go?

Also, it looks like token expires by default after 1 hour according to the docs. How should we handle this?

ivanstanev (commented):
Thank you for the contribution! I was wondering if you only meant to submit the `fsGroup` change?

PR #872, opened 2021-09-16
fix: allow configuring fsGroup and projected volumes for AWS EKS

ivanstanev:

### Notes for the reviewer

Resolves https://github.com/snyk/kubernetes-monitor/pull/871.
Added tests with Helm.

### DIFF
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -108,3 +108,18 @@ psp:
 
 # Override the excluded namespaces
 excludedNamespaces:
+
+# Allow specifying a fsGroup in the PodSpec securityContext:
+# spec:
+#   template:
+#     spec:
+#       securityContext:
+#         fsGroup: <-- here
+securityContext:
+  fsGroup:
+
+# A projected volume maps several existing volume sources into the same directory.
+# https://kubernetes.io/docs/concepts/storage/volumes/#projected
+volumes:
+  projected:
+    serviceAccountToken: false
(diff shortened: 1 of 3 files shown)


snyk-monitor/values.yaml
```diff
@@ -108,3 +108,18 @@ psp:
...
+#       securityContext:
+#         fsGroup: <-- here
+securityContext:
+  fsGroup:
+
+# A projected volume maps several existing volume sources into the same directory.
+# https://kubernetes.io/docs/concepts/storage/volumes/#projected
+volumes:
+  projected:
+    serviceAccountToken: false
```
minsiyang, line 125: This can overwrite the `serviveAccountToken` value and probably more subfields under there in the deployment? And the key is default to valid for 1 hr. Is that long enough? Is it for checking the role with AWS? 😵‍💫
https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html

minsiyang: And it doesn't require `volumeMount`?

ivanstanev: It would load the service account token into the container yep! And according to @tektite-io the AWS token is an OAuth token and it automatically gets refreshed.

ivanstanev: > And it doesn't require volumeMount?

Seems like it doesn't need to because the `serviceAccountToken` already gets mounted in the Pod. We are just projecting it to a different path!

minsiyang (approved):

PR #930, opened 2021-11-02
Feat/operator resources

ivanstanev:

### What this does

- allow configurable memory request/limit for the Operator
- paginate list() calls to the Kubernetes API 

### Additional information

- https://snyk.slack.com/archives/C01KTVA5S80/p1613644169019400

### DIFF
--- a/src/supervisor/watchers/handlers/deployment-config.ts
+++ b/src/supervisor/watchers/handlers/deployment-config.ts
@@ -1,6 +1,50 @@
+import { IncomingMessage } from 'http';
 import { deleteWorkload } from './workload';
 import { WorkloadKind } from '../../types';
-import { FALSY_WORKLOAD_NAME_MARKER, V1DeploymentConfig } from './types';
+import {
+  FALSY_WORKLOAD_NAME_MARKER,
+  V1DeploymentConfig,
+  V1DeploymentConfigList,
+} from './types';
+import { paginatedList } from './pagination';
+import { k8sApi } from '../../cluster';
+
+export async function paginatedDeploymentConfigList(
+  namespace: string,
+): Promise<{
+  response: IncomingMessage;
+  body: V1DeploymentConfigList;
+}> {
+  const v1DeploymentConfigList = new V1DeploymentConfigList();
+  v1DeploymentConfigList.apiVersion = 'apps.openshift.io/v1';
+  v1DeploymentConfigList.kind = 'DeploymentConfigList';
+  v1DeploymentConfigList.items = new Array<V1DeploymentConfig>();
+
+  return await paginatedList(
+    namespace,
+    v1DeploymentConfigList,
+    async (
+      namespace: string,
+      pretty?: string,
+      _allowWatchBookmarks?: boolean,
+      _continue?: string,
+      fieldSelector?: string,
+      labelSelector?: string,
+      limit?: number,
+    ) =>
+      k8sApi.customObjectsClient.listNamespacedCustomObject(
+        'apps.openshift.io',
+        'v1',
+        namespace,
+        'deploymentconfigs',
+        pretty,
+        _continue,
+        fieldSelector,
+        labelSelector,
+        limit,
+      ) as any,
+  );
+}
 
 export async function deploymentConfigWatchHandler(
   deploymentConfig: V1DeploymentConfig,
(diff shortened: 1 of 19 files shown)


src/supervisor/watchers/handlers/pod.ts
```diff
@@ -120,6 +124,22 @@ export function isPodReady(pod: V1Pod): boolean {
...
+}> {
+  const v1PodList = new V1PodList();
+  v1PodList.apiVersion = 'v1';
+  v1PodList.kind = 'PodList';
+  v1PodList.items = new Array<V1Pod>();
+
+  return await paginatedList(
+    namespace,
+    v1PodList,
+    k8sApi.coreClient.listNamespacedPod.bind(k8sApi.coreClient),
```
kat1906, line 139: what does this line do?

src/supervisor/watchers/handlers/pagination.ts
```diff
@@ -0,0 +1,92 @@
...
+        pretty,
+        allowWatchBookmarks,
+        continueToken,
+        fieldSelector,
+        labelSelector,
+        PAGE_SIZE,
+      );
+      incomingMessage = listCall.response;
+
+      if (Array.isArray(listCall.body.items)) {
```
kat1906, line 44: is it ok if it is not an array?

src/supervisor/watchers/handlers/pagination.ts
```diff
@@ -0,0 +1,92 @@
...
+        const trimmedItems = listCall.body.items.map((item: T) => {
+          delete item.metadata?.managedFields;
+          return item;
+        });
+        list.items.push(trimmedItems);
+      }
+
+      continueToken = listCall.body.metadata?._continue;
```
kat1906, line 52: what is a continue token? something that indicates there are more to list?

src/supervisor/watchers/handlers/pagination.ts
```diff
@@ -0,0 +1,92 @@
...
+          const seconds = calculateSleepSeconds(error.response);
+          await sleep(seconds);
+          continue;
+        default:
+          throw err;
+      }
+    }
+  }
+
+  if (!incomingMessage) {
```
kat1906, line 84: can this be moved to just below L42?

src/supervisor/watchers/handlers/pod.ts
ivanstanev: Just silencing the TS compiler/linter

`bind` ensures that if `listNamespacedPod` references `this` in its implementation, that `this` returns the right value

src/supervisor/watchers/handlers/pagination.ts
ivanstanev: This also implicitly covers an `undefined` or `null`

src/supervisor/watchers/handlers/pagination.ts
ivanstanev: Yes, something like "next page" starting from this token

src/supervisor/watchers/handlers/pagination.ts
ivanstanev: TS complains because there is a case where the first call returns 410 - we don't have an incoming message at that point

src/supervisor/watchers/handlers/pagination.ts
kat1906: maybe it could be renamed to be clearer? or is that it's actual name?

ivanstanev: Yeah the parameter is named `continue`, but I'm open to suggestions!

kat1906: ``shouldContinue = !!listCall.body.metadata?._continue;

ivanstanev: Ahh no, the return value of `_continue` actually a string, something like a token, hence it must be supplied as is. It's not really a flag/boolean!

ivanstanev: If you check out the documentation in VS Code, some of these parameters will start to make sense 😄 

Screenshot 2021-11-03 at 10 33 46

src/supervisor/watchers/handlers/deployment-config.ts
```diff
@@ -1,6 +1,50 @@
...
+        'apps.openshift.io',
+        'v1',
+        namespace,
+        'deploymentconfigs',
+        pretty,
+        _continue,
+        fieldSelector,
+        labelSelector,
+        limit,
+      ) as any,
```
ivanstanev, line 45: As any because the K8s API client returns `object` for the type of the items instead of type `T` 😞

kat1906 (approved):

PR #939, opened 2021-11-09
fix: trim workload metadata everywhere we read workloads

ivanstanev:

### What this does

I noticed a few other places where the workload metadata is not trimmed.
In the logs you could still see "managedFields" appearing, which is a pretty large object in some cases.
This change now ensures that we trim the metadata everywhere we receive a workload.

### DIFF
--- a/src/supervisor/watchers/handlers/pagination.ts
+++ b/src/supervisor/watchers/handlers/pagination.ts
@@ -5,6 +5,7 @@ import type {
   KubernetesObject,
 } from '@kubernetes/client-node';
 
+import { trimWorkloads } from './workload';
 import { calculateSleepSeconds } from '../../kuberenetes-api-wrappers';
 import { V1NamespacedList } from './types';
 import type { IRequestError } from '../../types';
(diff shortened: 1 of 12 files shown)


src/supervisor/watchers/handlers/pagination.ts
```diff
@@ -5,6 +5,7 @@ import type {
   KubernetesObject,
 } from '@kubernetes/client-node';
 
+import { trimItems } from './workload';
```
kat1906, line 8: where is this defined? I can' see it in the workload file

ivanstanev: https://github.com/snyk/kubernetes-monitor/pull/939/files#diff-497df571ce5ee886b45b30679d9f64cf189d762ca02bcfb2583178e2bfa23a8dR46

src/supervisor/watchers/handlers/pagination.ts
```diff
@@ -5,6 +5,7 @@ import type {
   KubernetesObject,
 } from '@kubernetes/client-node';
 
+import { trimWorkloads } from './workload';
```
kat1906, line 8: 👍

kat1906 (approved):

minsiyang (approved):

PR #975, opened 2021-12-22
Add init delay to liveness and readiness probes

stefansedich:

### What this does

We are attempting to use the monitor and in our environment are experiencing probes failing on startup. From our investigation we see the monitor pinning the CPU on startup which is causing the probes to fail and in turn the container to be killed before it starts up.

I was able to get things to startup sometimes by giving the container more CPU cores but even this was not reliable. Adding an init delay should hopefully resolve this and allow the container to start before it is nuked due to failing probes.

As the probes are not overly useful here I just went with a long init delay.

### DIFF
--- a/snyk-monitor/Chart.yaml
+++ b/snyk-monitor/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: A Helm chart for the Snyk Monitor
 name: snyk-monitor
-version: 0.1.0
+version: 0.1.1
 icon: https://res.cloudinary.com/snyk/image/upload/v1533761770/logo-1_wtob68.svg
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -141,10 +141,12 @@ spec:
               {{- toYaml . | nindent 14 }}
             {{- end }}
           livenessProbe:
+            initialDelaySeconds: 60
             exec:
               command:
               - "true"
           readinessProbe:
+            initialDelaySeconds: 60
             exec:
               command:
               - "true"


dkontorovskyy:
Hey @stefansedich! Thanks much for your submission!

Can you describe your issue in more detail, please?
We haven't seen similar behavior anywhere yet, so would like to understand your root cause.

Thanks :)

stefansedich:
> Hey @stefansedich! Thanks much for your submission!
> 
> Can you describe your issue in more detail, please?
> We haven't seen similar behavior anywhere yet, so would like to understand your root cause.
> 
> Thanks :) 

We see the container pin all CPU we give it on start-up causing the probes to fail which in turn causes the container to be terminated.

dkontorovskyy:
I'm curious why this solves your issue since liveness and readiness are only returning true 🤔 
Does your node has enough CPU allocated?

stefansedich:
> I'm curious why this solves your issue since liveness and readiness are only returning true 🤔 Does your node has enough CPU allocated?

It is odd, but I was able to get the agent to start giving it 4 cores which it was pinning on startup for some time.

stefansedich:
@dkontorovskyy do you have any further ideas here? From 1-4 cores all I see on startup is the CPU pinned and the probes start to fail, I have experimented with longer init delays and probe timeouts with little success.

stefansedich:
@dkontorovskyy FYI we upgraded to 1.80.0 which appeared to have CPU related fixes, and we dialed workers down to 5 and appear to be having more success getting the agent to start.

dkontorovskyy:
@stefansedich It's great news! Glad that it worked!

stefansedich:
> @stefansedich It's great news! Glad that it worked!

Sadly that was short lived :( looks like we hit the same issues overnight, I have jumped down to 1 worker and assigned 2 cores to the pod to see if that helps, do you have any more suggestions? should we reduce the compression settings too?

PR #988, opened 2022-01-12
RUN-1968 fix: sanitize workload metadata

minsiyang:

### What this does

Sanitize workload metadata before sending it to K-U

### DIFF
--- a/test/fixtures/sidecar-containers/deployment.yaml
+++ b/test/fixtures/sidecar-containers/deployment.yaml
@@ -48,6 +48,12 @@ spec:
           successThreshold: 1
           timeoutSeconds: 5
         name: hello-world
+        args: [something]
+        env: 
+          - name: NODE_EXTRA_CA_CERTS
+            value: "important info"
+        command:
+          - "do something"
         ports:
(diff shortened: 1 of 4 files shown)


test/fixtures/sidecar-containers/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
           successThreshold: 1
           timeoutSeconds: 5
         name: hello-world
+        args: [something]
```
dkontorovskyy, line 51: Let's add `env` also?

Should we have an explicit test to validate that this key is not being sent?

dkontorovskyy:
AMAZING 👑

test/fixtures/sidecar-containers/deployment.yaml
minsiyang: the unit test of `metadata-extractor.spec.ts` should validate that?

minsiyang: I see what you mean now, sorry. I will add a test now 😅

PR #990, opened 2022-01-17
fix: retry on network errors in K8s API informers

ivanstanev:

### What this does

Add several other errors to the list of retryable errors for when the informer faces a network issue. This should make reconnecting to the K8s API more robust as these network errors are normally only temporary.

### DIFF
--- a/src/supervisor/watchers/handlers/index.ts
+++ b/src/supervisor/watchers/handlers/index.ts
@@ -30,7 +30,7 @@ import {
 import { k8sApi, kubeConfig } from '../../cluster';
 import * as kubernetesApiWrappers from '../../kuberenetes-api-wrappers';
 import { IWorkloadWatchMetadata, FALSY_WORKLOAD_NAME_MARKER } from './types';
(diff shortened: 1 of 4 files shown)


src/supervisor/watchers/handlers/index.ts
```diff
@@ -211,18 +212,22 @@ export async function setupInformer(
...
-        {},
-        `informer ${ECONNRESET_ERROR_CODE} occurred, restarting informer`,
+        logContext,
+        `informer ${code} occurred, restarting informer`,
       );
 
       // Restart informer after 1sec
       setTimeout(async () => {
         await informer.start();
       }, 1000);
```
minsiyang, line 225: So this returns a Promise. It doesn't throw err? If this fails, then do we retry?

ivanstanev: The `.start` function doesn't itself throw. Instead we'll get a event in this same error handler, so it will enter this piece of code again

minsiyang (approved):

PR #993, opened 2022-01-17
fix: pull OCI archives instead of Docker archives

ivanstanev:

### What this does

For some image archives this results in much smaller pulled images as it allows to directly download compressed layers.
Also, it reduces the need to convert between different formats on the fly, which skopeo had to do for us.

### DIFF
--- a/src/scanner/index.ts
+++ b/src/scanner/index.ts
@@ -17,7 +17,11 @@ import {
   ILocalWorkloadLocator,
   Telemetry,
 } from '../transmitter/types';
-import { IPullableImage, IScanImage } from './images/types';
+import {
+  IPullableImage,
+  IScanImage,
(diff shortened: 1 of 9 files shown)


src/scanner/index.ts
```diff
@@ -74,42 +78,37 @@ export async function sendDeleteWorkloadRequest(
...
-      accum[meta.imageName] = {
-        imageWithDigest: digest && `${imageName}@${digest}`,
-        imageName: meta.imageName, // Image name with tag
-      };
-
-      return accum;
-    },
-    {},
-  );
+  const uniqueImages = workloadMetadata.reduce((accum, meta) => {
```
ivanstanev, line 81: Removed the type since it's now assigned to the `accumulator`, but then prettier reformatted everything

PR #994, opened 2022-01-21
[RUN-1518] Add support for CronJobs in batch/v1 API

kat1906:

### What this does

The CronJob workload that snyk-monitor supports is the batch/v1beta1 API which has been around for a long time. Starting version 1.21 of Kubernetes this API has graduated and is now GA under batch/v1. This PR updates the monitor to use that version.

### DIFF
--- a/src/supervisor/watchers/handlers/index.ts
+++ b/src/supervisor/watchers/handlers/index.ts
@@ -69,7 +69,7 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
       paginatedReplicationControllerList(namespace),
   },
   [WorkloadKind.CronJob]: {
-    endpoint: '/apis/batch/v1beta1/watch/namespaces/{namespace}/cronjobs',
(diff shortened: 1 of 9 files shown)


src/supervisor/watchers/handlers/index.ts
```diff
@@ -69,7 +69,7 @@ const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
       paginatedReplicationControllerList(namespace),
   },
   [WorkloadKind.CronJob]: {
```
ivanstanev, line 71: I wonder if we need to introduce a new workload kind - CronJobV1, and the existing one becomes CronJobV1Beta1. This way we can run both readers side by side, otherwise we will only be able to detect one but not the other

kat1906: but this is based on kind which is `CronJob` for both!

kat1906: the v1 reader will work for v1 and v1beta1 CronJobs (verified by fixture in integration test)

minsiyang (approved):

PR #998, opened 2022-02-03
Feat/cluster informers

ivanstanev:

### What this does

Significantly reduce the number of watches to the Kubernetes API.
We now set up only up to 9 watches/informers with the Kubernetes API for the whole cluster instead of up to 9 watches per namespace.
For really large clusters this is a great improvement - instead of setting thousands of watches, we only set up to 9, regardless of the cluster size.


src/supervisor/watchers/handlers/index.ts
```diff
@@ -292,12 +430,13 @@ export async function setupInformer(
...
-    namespace,
-  );
+  const endpoint = shouldWatchCluster
+    ? workloadMetadata.clusterEndpoint
+    : workloadMetadata.namespacedEndpoint.replace('{namespace}', namespace);
 
-  const listMethod = workloadMetadata.namespacedListFactory(namespace);
+  const listMethod = shouldWatchCluster
+    ? workloadMetadata.clusterListFactory()
+    : workloadMetadata.namespacedListFactory(namespace);
```
minsiyang, line 439: Do we want to know whether the error comes from namespacedListFactory or clusterListFactory?

src/supervisor/watchers/handlers/index.ts
```diff
@@ -11,32 +11,59 @@ import {
...
-import { daemonSetWatchHandler, paginatedDaemonSetList } from './daemon-set';
-import { deploymentWatchHandler, paginatedDeploymentList } from './deployment';
-import { jobWatchHandler, paginatedJobList } from './job';
-import { paginatedReplicaSetList, replicaSetWatchHandler } from './replica-set';
 import {
-  paginatedReplicationControllerList,
+  daemonSetWatchHandler,
+  paginatedClusterDaemonSetList,
+  paginatedNamespacedDaemonSetList,
+} from './daemon-set';
```
kat1906, line 31: maybe the exports could be changed to import just 1 thing e.g. `daemonSet` and then could call the methods from that e.g. `demonSet.watchHandler`, `demonSet.paginatedSetList` and `demonSet.paginatedNamespacedSetList` ... also could drop the paginated all together and just have comment in the file?

src/supervisor/watchers/handlers/pagination.ts
```diff
@@ -99,3 +99,87 @@ export async function paginatedNamespacedList<
...
   };
 }
+
+/**
+ * This function ensures that when listing workloads from the Kubernetes API, they are paginated in batches of 100.
+ * The workloads collected are additionally trimmed to contain only the relevant data for vulnerability analysis.
+ * The combination of both listing and trimming ensures we reduce our memory footprint and prevent overloading the API server.
+ */
+export async function paginatedClusterList<
+  T extends KubernetesObject & Partial<{ status: unknown; spec: unknown }>,
```
minsiyang, line 109: This function is almost identical to the paginatedNamespacedList function, can we make `namespace` parameter a optional parameter, and the reuse the paginatedNamespacedList function somehow? Unless we are planning on getting rid of it in the future?

src/supervisor/watchers/handlers/index.ts
```diff
@@ -58,97 +85,133 @@ import { RETRYABLE_NETWORK_ERRORS } from '../types';
...
     },
-    listFactory: (namespace) => () => paginatedDeploymentConfigList(namespace),
+    clusterListFactory: () => () => paginatedClusterDeploymentConfigList(),
+    namespacedListFactory: (namespace) => () =>
+      paginatedNamespacedDeploymentConfigList(namespace),
   },
 };
 
-async function isSupportedWorkload(
+export const WATCH_WHOLE_CLUSTER = '';
```
kat1906, line 198: I don't understand what this const does?

kat1906: I guess to decide if the whole cluster watch is set up ... is empty string the safest way to do this?
```suggestion
export const WATCH_WHOLE_CLUSTER = 'WATCH_WHOLE_CLUSTER';
```

src/supervisor/watchers/index.ts
```diff
@@ -60,52 +57,18 @@ export function isExcludedNamespace(namespace: string): boolean {
...
-        return;
-      }
-
-      await setupWatchesForNamespace(namespace);
-    } catch (err) {
-      logger.error({ err, namespace }, 'error handling a namespace event');
-      return;
+      await setupInformer(WATCH_WHOLE_CLUSTER, workloadKind);
+    } catch (error) {
+      logger.warn({ workloadKind }, 'could not setup workload watch, skipping');
```
minsiyang, line 69: Do you want to know this error is coming from the setting up informer for the cluster?

src/supervisor/watchers/handlers/pagination.ts
ivanstanev: Unfortunately not possible to re-use because the order of parameters is different in these functions!

src/supervisor/watchers/handlers/index.ts
```diff
@@ -319,7 +133,88 @@ export async function setupInformer(
...
+    loggedListMethod,
+  );
+
+  informer.on(ERROR, (err) => {
+    // Types from client library insists that callback is of type KubernetesObject
+    const code = (err as any).code || '';
+    if (RETRYABLE_NETWORK_ERRORS.includes(code)) {
+      logger.debug(
+        logContext,
+        `informer ${code} occurred, restarting informer`,
```
kat1906, line 188: should we be using string interpolation in error messages?

ivanstanev: good point, we shouldn't!

src/supervisor/watchers/handlers/informer-config.ts
```diff
@@ -0,0 +1,149 @@
...
+ * The map is keyed by the "WorkloadKind" -- the type of resource we want to watch.
+ * Legal verbs for the "handlers" are pulled from '@kubernetes/client-node'. You can
+ * set a different handler for every verb.
+ * (e.g. ADD-ed workloads are processed differently than DELETE-d ones)
+ *
+ * The "listFunc" is a callback used by the kubernetes-client to grab the watched resource
+ * whenever Kubernetes fires a "workload changed" event and it uses the result to figure out
+ * if the workload actually changed (by inspecting the resourceVersion).
+ */
+export const workloadWatchMetadata: Readonly<IWorkloadWatchMetadata> = {
```
kat1906, line 33: I _feel_ like this object could be even more streamlined, there is a lot of repeated pattern. could you have a function that passes in the imported e.g. `pod` and returns the config? the names of the handlers would have to be standardised e.g. `podWatchHandler` to `watchHandler` - this would make adding one even easier!

src/supervisor/watchers/handlers/pagination.ts
kat1906: you could wrap the params in an object?

ivanstanev: I don't think it's possible still to call the function, I cannot `await listFn(...params)`

kat1906 (approved):

PR #1022, opened 2022-03-09
[RUN-2130] Feat/sysdig healthcheck

ivanstanev:

### What this does

Introduce periodic logging healthcheck.
This helps us to better understand the status of the snyk-monitor, for example, whether accessing the Sysdig integration is possible.


src/data-scraper/index.ts
```diff
@@ -7,23 +7,59 @@ import { IRuntimeImagesResponse } from '../transmitter/types';
...
+    headers: {
+      authorization: header,
+    },
+    timeout: 10_000,
+  };
+
+  const limit: number = 1;
+  const cursor: string = '';
+  const { response } = await retryRequest(
+    'get',
```
dkontorovskyy, line 45: If Sysdig supports HEAD requests, we can try that, but this is nit 🤗

ivanstanev: After discussing it, the current approach seems to be the recommended one

src/data-scraper/index.ts
```diff
@@ -7,23 +7,59 @@ import { IRuntimeImagesResponse } from '../transmitter/types';
...
+const httpsAgent = new HttpsAgent({
+  keepAlive: true,
+  // We agreed with Sysdig to skip TLS certificates validation for HTTPS connection.
+  rejectUnauthorized: false,
+});
+
+function getSysdigUrl(): string {
+  const base: string = config.SYSDIG_ENDPOINT;
+  const url: string = base + '/v1/runtimeimages';
+  return url;
```
kat1906, line 19: nit: you could interpolate like is done for auth header?

src/data-scraper/index.ts
```diff
@@ -7,23 +7,59 @@ import { IRuntimeImagesResponse } from '../transmitter/types';
...
+
+function getSysdigAuthHeader(): string {
+  return `Bearer ${config.SYSDIG_TOKEN}`;
+}
+
 function isSuccessStatusCode(statusCode: number | undefined): boolean {
   return statusCode !== undefined && statusCode >= 200 && statusCode < 300;
 }
 
+/** NOTE: This function can throw, so the caller should handle errors. */
```
kat1906, line 30: nit: do we have to throw?

src/data-scraper/index.ts
```diff
@@ -7,23 +7,59 @@ import { IRuntimeImagesResponse } from '../transmitter/types';
...
+  const limit: number = 1;
+  const cursor: string = '';
+  const { response } = await retryRequest(
+    'get',
+    `${url}?limit=${limit}&cursor=${cursor}`,
+    {},
+    reqOptions,
+  );
+  if (!isSuccessStatusCode(response.statusCode)) {
+    throw new Error(`${response.statusCode} ${response.statusMessage}`);
```
minsiyang, line 51: Should we add logging to this error, so we know what is happening?

minsiyang: never mind, just saw that it was logging in the healthcheck.ts.

src/index.ts
```diff
@@ -69,20 +86,6 @@ setImmediate(async function setUpAndMonitor(): Promise<void> {
...
-          { error },
-          'an error occurred while scraping runtime data',
-        );
-      }
-    }, interval).unref();
-  } else {
-    logger.info({}, 'Sysdig integration not detected');
-  }
+  await setupSysdigIntegration();
+  await setupHealthCheck();
```
minsiyang, line 90: Should the healtherCheck be set up inside the `setupSysdigIntegration`?

ivanstanev: I wanted to make it a bit more generic, so it's not only about Sysdig but also about other factors in the future!

src/data-scraper/index.ts
ivanstanev: Not necessary, but I would like the healthcheck to log if we detect an exception (the throwing is not only because of us, but also it could be the HTTP client returning a network error like ECONNRESET)

kat1906 (approved):

minsiyang (approved):

PR #1026, opened 2022-03-10
fix: track image state to avoid sending data for deleted workloads

ivanstanev:

### What this does

Small bugfix to how we track state. Previously we were not actually deleting keys from the cache at all. This could result in a race condition where we would still send scan results to the upstream even though the workload was already deleted.

We used to set the key in the cache by "image id" but try to delete by "image name"

### DIFF
--- a/src/supervisor/watchers/handlers/pod.ts
+++ b/src/supervisor/watchers/handlers/pod.ts
@@ -95,12 +95,19 @@ async function handleReadyPod(workloadMetadata: IWorkload[]): Promise<void> {
   for (const workload of workloadMetadata) {
     const scanned = await getWorkloadImageAlreadyScanned(
       workload,
-      workload.imageId,
+      workload.imageName,
     );
-    if (scanned !== undefined) {
+    // ImageID contains the resolved image digest.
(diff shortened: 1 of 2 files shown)


src/supervisor/watchers/handlers/pod.ts
```diff
@@ -95,12 +95,19 @@ async function handleReadyPod(workloadMetadata: IWorkload[]): Promise<void> {
...
-    if (scanned !== undefined) {
+    // ImageID contains the resolved image digest.
+    // ImageName may contain a tag. The image behind this tag can be mutated and can change over time.
+    // We need to compare on ImageID which will reliably tell us if the image has changed.
+    if (scanned === workload.imageId) {
       continue;
     }
-    await setWorkloadImageAlreadyScanned(workload, workload.imageId, ''); // empty string takes zero bytes and is !== undefined
+    await setWorkloadImageAlreadyScanned(
+      workload,
```
minsiyang, line 107: nit: Should we update the codebase wherever referring to `workload.imageId` or `imageId` that actually mean image name to imageName. It was quite confusing 😄

minsiyang (approved):

PR #1035, opened 2022-03-21
[RUN-2125] Feat/delete workloads from scan queue

ivanstanev:

### What this does

1. remove workloads from scan queue when they are deleted in K8s

Up until now the queue of workloads to scan would keep growing indefinitely as we only appended new workloads to it.
However, once workloads are deleted from Kubernetes, they no longer need to be in the queue.

This didn't cause any issues with trying to process deleted workloads, because the queue handler would detect that the workload is gone and continue processing other items. However, it created alerts that the queue grows indefinitely and caused the snyk-monitor to process some work that was never going to succeed.

2. control with flag if failed or complete jobs should be processed

The undocumented SKIP_K8S_JOBS flag can now control whether failed or complete Jobs should be processed or removed from the scan queue.
This behaviour is not enabled by default because we are trying to work around a limitation where unique jobs using the exact same image are created many many times.
In the future this will be improved by adding image scan results caching.


kat1906 (approved):

src/supervisor/watchers/handlers/cron-job.ts
```diff
@@ -117,6 +118,7 @@ export async function cronJobWatchHandler(
           .filter((container) => container.image !== undefined)
           .map((container) => container.image!),
       }),
+      deleteWorkloadFromScanQueue(workloadAlreadyScanned),
```
dkontorovskyy, line 121: can `deleteWorkloadFromScanQueue` be used somewhere after we are calling this function? Or because there is a different process for different types of workloads we need to have this step inside each function?

ivanstanev: The handlers are the entrypoints, there isn't a generic function at the moment where we can simply put this code to allow re-use. I know it can be a bit error prone! I think there is a way to generalize it but I wonder if it's a bit too premature

dkontorovskyy (approved):

config.default.json
```diff
@@ -11,9 +11,9 @@
     "MAX_SIZE": 10000,
     "MAX_AGE_MS": 60000
   },
-  "WORKERS_COUNT": 10,
+  "WORKERS_COUNT": 5,
   "REQUEST_QUEUE_LENGTH": 2,
-  "QUEUE_LENGTH_LOG_FREQUENCY_MINUTES": 15,
+  "QUEUE_LENGTH_LOG_FREQUENCY_MINUTES": 1,
```
kat1906, line 16: is this intentional?

ivanstanev: Yes, it's generating a debug log so for almost everyone it would not have any effect!
I find that 15 mins is too infrequent especially if you are looking at the snyk-monitor running in the cluster, it takes a long time to debug the queue growth

dkontorovskyy (approved):

PR #1039, opened 2022-03-30
Feat/alpine base image

ivanstanev:

### What this does

Load correct image in integration tests for OpenShift.

### DIFF
--- a/scripts/circleci-jobs/setup-integration-tests.py
+++ b/scripts/circleci-jobs/setup-integration-tests.py
@@ -14,7 +14,7 @@ if circle_branch == "staging":
     image_tag = "staging-candidate"
 else:
     image_tag = "discardable"
-circle_sha1 = os.getenv("CIRCLE_SHA1")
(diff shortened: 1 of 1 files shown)


scripts/circleci-jobs/setup-integration-tests.py
```diff
@@ -14,7 +14,7 @@
     image_tag = "staging-candidate"
 else:
     image_tag = "discardable"
-circle_sha1 = os.getenv("CIRCLE_SHA1")
+circle_sha1 = os.getenv("CIRCLE_SHA1") if image_tag_suffix == '' else os.getenv("CIRCLE_SHA1")[0:8]
```
minsiyang, line 17: I don't quite understand what the `if` statement here is doing? So if the `image_tag_suffix` is an empty string, the `circle_sha1` would be the first 8 characters of the `"CIRCLE_SHA1"`?

ivanstanev: If the suffix is empty, we get the full SHA, otherwise if we have a suffix then we pick only the 8 chars from the SHA

kat1906 (approved):

minsiyang (approved):

PR #1050, opened 2022-04-07
RUN-2198 chore: add more logs for debug purpose

minsiyang:

### What this does

* capture error message from SKOPEO better
* add more logs to understand why some workloads don't produce scan result

### Screenshots

### DIFF
--- a/src/supervisor/metadata-extractor.ts
+++ b/src/supervisor/metadata-extractor.ts
@@ -30,6 +30,7 @@ export function buildImageMetadata(
     delete container.args;
     delete container.env;
     delete container.command;
+    //! would container.envFrom also include sensitive data?
     containerNameToSpec[container.name] = container;
   }
 
@@ -41,6 +42,17 @@ export function buildImageMetadata(
   const images: IWorkload[] = [];
   for (const containerStatus of containerStatuses) {
     if (!(containerStatus.name in containerNameToSpec)) {
+      logger.debug(
+        {
+          workloadName: workloadMeta.objectMeta.name,
+          workloadType: workloadMeta.kind,
+          workloadNamespace: workloadMeta.objectMeta.namespace,
+          containerName: containerStatus.name,
+          image: containerStatus.image,
+          imageId: containerStatus.imageID,
+        },
+        'container name is not in containerNameToSpec lists',
+      );
       continue;
     }
     images.push({
@@ -143,6 +155,16 @@ export async function buildMetadataForWorkload(
   const isAssociatedWithParent = isPodAssociatedWithParent(pod);
 
   if (!pod.metadata || pod.metadata.namespace === undefined || !pod.spec) {
+    const logContext = {
+      workloadName: pod.metadata?.name,
+      workloadType: pod.kind,
+      workloadNamespace: pod.metadata?.namespace,
+      clusterName: pod.metadata?.clusterName,
+    };
+    logger.debug(
+      logContext,
+      'cannot build metadata for workload without pod information',
+    );
     // Some required parameters are missing, we cannot process further
     return undefined;
   }
--- a/src/supervisor/watchers/handlers/pod.ts
+++ b/src/supervisor/watchers/handlers/pod.ts
@@ -38,8 +38,16 @@ export async function handleReadyPod(
     // ImageName may contain a tag. The image behind this tag can be mutated and can change over time.
     // We need to compare on ImageID which will reliably tell us if the image has changed.
     if (scanned === workload.imageId) {
+      logger.debug(
+        { workloadToScan, imageId: workload.imageId },
+        'image already scanned',
+      );
       continue;
     }
+    logger.debug(
+      { workloadToScan, imageId: workload.imageId },
+      'image has not been scanned',
+    );
     await setWorkloadImageAlreadyScanned(
       workload,
       workload.imageName,
@@ -59,16 +67,28 @@ export async function handleReadyPod(
 }
 
 export function isPodReady(pod: V1Pod): boolean {
+  const podStatus = pod.status !== undefined;
+  const podStatusPhase = pod.status?.phase === PodPhase.Running;
+  const podContainerStatuses = pod.status?.containerStatuses !== undefined;
+  const containerReadyStatuses = pod.status?.containerStatuses?.some(
+    (container) =>
+      container.state !== undefined &&
+      (container.state.running !== undefined ||
+        container.state.waiting !== undefined),
+  ) as boolean;
+
+  const logContext = {
+    podStatus,
+    podStatusPhase,
+    podContainerStatuses,
+    containerReadyStatuses,
+  };
+  logger.debug(logContext, 'checking to see if pod is ready to process');
   return (
-    pod.status !== undefined &&
-    pod.status.phase === PodPhase.Running &&
-    pod.status.containerStatuses !== undefined &&
-    pod.status.containerStatuses.some(
-      (container) =>
-        container.state !== undefined &&
-        (container.state.running !== undefined ||
-          container.state.waiting !== undefined),
-    )
+    podStatus &&
+    podStatusPhase &&
+    podContainerStatuses &&
+    containerReadyStatuses
   );
 }
 
--- a/src/supervisor/workload-sanitization.ts
+++ b/src/supervisor/workload-sanitization.ts
@@ -1,4 +1,5 @@
 import { KubernetesObject, V1ObjectMeta } from '@kubernetes/client-node';
+import { logger } from '../common/logger';
 
 export function trimWorkloads<
   T extends KubernetesObject & Partial<{ status: unknown; spec: unknown }>,
@@ -14,6 +15,7 @@ export function trimWorkloads<
 export function trimWorkload<
   T extends KubernetesObject & Partial<{ status: unknown; spec: unknown }>,
 >(workload: T): T & { metadata: V1ObjectMeta } {
+  logger.debug(workload.metadata, 'workload metadata state before trimming');
   return {
     apiVersion: workload.apiVersion,
     kind: workload.kind,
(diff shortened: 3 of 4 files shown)


src/supervisor/watchers/handlers/pod.ts
```diff
@@ -38,6 +38,10 @@ export async function handleReadyPod(
     // ImageName may contain a tag. The image behind this tag can be mutated and can change over time.
     // We need to compare on ImageID which will reliably tell us if the image has changed.
     if (scanned === workload.imageId) {
+      logger.debug(
+        { scanned: scanned, imageId: workload.imageId },
+        'image already scanned',
+      );
       continue;
     }
     await setWorkloadImageAlreadyScanned(
```
minsiyang, line 47: Should we push the `workload` to `workloadToScan` first before `setWorkloadImageAlreadyScanned`? Or does it not make much time difference?

src/supervisor/metadata-extractor.ts
```diff
@@ -30,6 +30,7 @@ export function buildImageMetadata(
     delete container.args;
     delete container.env;
     delete container.command;
+    //! would container.envFrom also include sensitive data?
```
ivanstanev, line 33: It shouldn't, it contains only a name to a secret or a configmap - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#envfromsource-v1-core

src/supervisor/metadata-extractor.ts
```diff
@@ -41,6 +42,10 @@ export function buildImageMetadata(
   const images: IWorkload[] = [];
   for (const containerStatus of containerStatuses) {
     if (!(containerStatus.name in containerNameToSpec)) {
+      logger.debug(
+        { containerName: containerStatus.name },
```
ivanstanev, line 46: Is it possible to add more context, otherwise how can we distinguish which container comes from where? For example, adding the workload name, type, and namespace!

src/supervisor/metadata-extractor.ts
```diff
@@ -54,6 +59,7 @@ export function buildImageMetadata(
       specAnnotations: specMeta.annotations || {},
       containerName: containerStatus.name,
       imageName: containerNameToSpec[containerStatus.name].image,
+      //! why do we use containerNameToSpec rather than containerNameToStatus here?
```
ivanstanev, line 62: The spec contains the unprocessed data, as the user has requested it when applying the deployment. The status contains the processed data about the running pod, hence imageID shows there (which is the resolved data after pulling the image).

I had a look at a few examples and in both the spec and in the status the image name doesn't seem to change. Not sure why we chose one over the other, probably because of preferring the assurance that the spec won't be modified

src/supervisor/metadata-extractor.ts
```diff
@@ -143,6 +149,14 @@ export async function buildMetadataForWorkload(
   const isAssociatedWithParent = isPodAssociatedWithParent(pod);
 
   if (!pod.metadata || pod.metadata.namespace === undefined || !pod.spec) {
+    const logContext = {
+      podMetadata: pod.metadata,
+      podSpec: pod.spec,
```
ivanstanev, line 154: Both of these can hold a lot of data. Instead could we log only selected fields that can help us with debugging? For example, the workload name, type and namespace, and its images

src/supervisor/watchers/handlers/pod.ts
ivanstanev: Time difference in what sense? The `await` is just future-proofing the function in case we store this data in an external cache 😄 Since it now sets data in memory it's pretty much as fast as it gets

src/supervisor/watchers/handlers/pod.ts
```diff
@@ -59,6 +67,18 @@ export async function handleReadyPod(
...
+    podStatusPhase: pod.status?.phase === PodPhase.Running,
+    podContainerStatues: pod.status?.containerStatuses !== undefined,
+    containerReadyStatues: pod.status?.containerStatuses?.some(
+      (container) =>
+        container.state !== undefined &&
+        (container.state.running !== undefined ||
+          container.state.waiting !== undefined),
+    ),
+  };
+  logger.debug(logContext, 'checking to see if pod is ready to process');
```
ivanstanev, line 81:
```typescript
const podStatus = pod.status !== undefined;
const podStatusPhase = pod.status?.phase === PodPhase.Running;
const podContainerStatuses = pod.status?.containerStatuses !== undefined;
const containerReadyStatuses = pod.status?.containerStatuses?.some((container) =>
  container.state !== undefined &&
  (container.state.running !== undefined ||
    container.state.waiting !== undefined),
);
const logContext = {
  podStatus, podStatusPhase, podContainerStatuses, containerReadyStatuses,
};
logger.debug(logContext, 'checking to see if pod is ready to process');
return podStatus && podStatusPhase && podContainerStatuses && containerReadyStatuses;
```

src/supervisor/watchers/handlers/pod.ts
```diff
@@ -59,6 +67,18 @@ export async function handleReadyPod(
 }
 
 export function isPodReady(pod: V1Pod): boolean {
+  const logContext = {
+    podStatus: pod.status !== undefined,
+    podStatusPhase: pod.status?.phase === PodPhase.Running,
+    podContainerStatues: pod.status?.containerStatuses !== undefined,
+    containerReadyStatues: pod.status?.containerStatuses?.some(
```
ivanstanev, line 74: Small typo: `Statuses`

src/supervisor/workload-sanitization.ts
```diff
@@ -24,6 +26,17 @@ export function trimWorkload<
...
+    name: metadata?.name,
+    namespace: metadata?.namespace,
+    annotations: metadata?.annotations,
+    labels: metadata?.labels,
+    ownerReferences: metadata?.ownerReferences,
+    uid: metadata?.uid,
+    resourceVersion: metadata?.resourceVersion,
+    generation: metadata?.generation,
+  };
+  logger.debug(logContext, 'workload metadata state after trimming');
```
ivanstanev, line 39: This seems to be repeating, how about:
```typescript
const trimmedMetadata = {
    name: metadata?.name,
    namespace: metadata?.namespace,
    annotations: metadata?.annotations,
    labels: metadata?.labels,
    ownerReferences: metadata?.ownerReferences,
    uid: metadata?.uid,
    resourceVersion: metadata?.resourceVersion,
    generation: metadata?.generation,
  };
  logger.debug(trimmedMetadata, 'workload metadata state after trimming');
  return trimmedMetadata;
```

src/supervisor/workload-sanitization.ts
```diff
@@ -14,6 +15,7 @@ export function trimWorkloads<
 export function trimWorkload<
   T extends KubernetesObject & Partial<{ status: unknown; spec: unknown }>,
 >(workload: T): T & { metadata: V1ObjectMeta } {
+  logger.debug(workload, 'workload state before trimming');
```
ivanstanev, line 18: After further considering it, we probably don't want to log the full workload. It will be extremely noisy! I think the logging we do in `trimMetadata()` should be enough

src/supervisor/metadata-extractor.ts
minsiyang: yes, but we also don't need to collect this information as well, right?

src/supervisor/metadata-extractor.ts
minsiyang: yes, it should produce the same result by looking at the `V1Container` spec and `V1ContainerStatus` spec. 😄

src/supervisor/watchers/handlers/pod.ts
minsiyang: ok. well, I don't know what the time difference would be. But in the case of the order, I thought it should be pushed to the  `workloadToScan`, and then marked the workload as `setWorkloadImageAlreadyScanned`. I guess the scanned status and the `workloadToScan` is totally two different thing, so the order here doesn't really matter :)

src/supervisor/workload-sanitization.ts
minsiyang: Or we can log the `Metadata` before trimming?

src/supervisor/metadata-extractor.ts
ivanstanev: Right now we don't have use for it, so we indeed don't need to collect it

src/supervisor/watchers/handlers/pod.ts
```diff
@@ -59,16 +67,28 @@ export async function handleReadyPod(
...
 export function isPodReady(pod: V1Pod): boolean {
+  const podStatus = pod.status !== undefined;
+  const podStatusPhase = pod.status?.phase === PodPhase.Running;
+  const podContainerStatuses = pod.status?.containerStatuses !== undefined;
+  const containerReadyStatuses = pod.status?.containerStatuses?.some(
+    (container) =>
+      container.state !== undefined &&
+      (container.state.running !== undefined ||
+        container.state.waiting !== undefined),
+  ) as boolean;
```
ivanstanev, line 78: Why the `as boolean`?

ivanstanev (approved):

src/supervisor/watchers/handlers/pod.ts
ivanstanev: In this case the order probably doesn't matter because the actions happen one after another and should produce identical results.

If these were operations that could yield the event loop then the order would matter. If we pushed to the queue and tried to store the image to the cache, but that yielded the loop and caused a similar task for the same workload to be run again, then we could risk pushing the same workload to the queue twice. https://snyk.io/blog/nodejs-how-even-quick-async-functions-can-block-the-event-loop-starve-io/ (the "👋 Micro-Tasks" section) Since our `await` doesn't do that here, the order won't matter 😄

src/supervisor/watchers/handlers/pod.ts
minsiyang: because the return type was `boolean | undefine`

src/supervisor/metadata-extractor.ts
minsiyang: Would we have any use of it in the future? Client's secret has nothing to do with us detecting their vulns?

PR #1055, opened 2022-04-20
[RUN-2187] Make keepalive configurable

kat1906:

### What this does

A customer is having network level issues. We want to rule out the new addition of keep-alive headers as being the problem so are making this configurable.


snyk-operator-certified/helm-charts/snyk-monitor/values.yaml
```diff
@@ -83,6 +83,7 @@ limits:
 http_proxy:
 https_proxy:
 no_proxy:
+use_keepalive:
```
ivanstanev, line 86: I think we want this to be `true` otherwise I think keep alive will be disabled by default

test/common/config.spec.ts
```diff
@@ -74,6 +74,7 @@ describe('extractNamespaceName()', () => {
     expect(config.HTTPS_PROXY).toBeUndefined();
     expect(config.HTTP_PROXY).toBeUndefined();
     expect(config.NO_PROXY).toBeUndefined();
+    expect(config.USE_KEEPALIVE).toEqual(true);
```
ivanstanev, line 77: WDYT of adding a similar check to the integration test `kubernetes.spec.ts` where we verify at runtime the env vars, volumes, securityContext, etc.

ivanstanev (approved):

minsiyang (approved):

PR #1059, opened 2022-04-22
fix: operator upgrade tests

kat1906:

### What this does

Operator upgrade tests are currently broken. This change reverts the changes in the "Upgrade Operator and check that snyk-monitor also upgraded" step and also checks for the operator tag without -ubi suffix

### DIFF
--- a/.circleci/config/jobs/operator_upgrade_tests.yml
+++ b/.circleci/config/jobs/operator_upgrade_tests.yml
@@ -175,21 +175,18 @@ steps:
         SLEEP_SECONDS_BETWEEN_ATTEMPTS=5
         # total = 10 minutes wait time
 
-        EXPECTED_TAG="${LATEST_TAG}-ubi8"
         # Periodically poll if the snyk-monitor has upgraded
         for (( attempt=1; attempt<ATTEMPTS; attempt++))
         do
-          # Grab the tag of the snyk-monitor container image.
-          SNYK_MONITOR_POD=$(kubectl get pods -n snyk-monitor --no-headers | grep "snyk-monitor")
-          if [[ ! -z "${SNYK_MONITOR_POD}" ]]; then
-            VERSION=$(echo "${SNYK_MONITOR_POD}" | \
-            awk 'END { print $1 }' | \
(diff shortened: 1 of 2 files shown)


.circleci/config/jobs/operator_upgrade_tests.yml
```diff
@@ -175,21 +175,18 @@ steps:
         SLEEP_SECONDS_BETWEEN_ATTEMPTS=5
         # total = 10 minutes wait time
 
-        EXPECTED_TAG="${LATEST_TAG}-ubi8"
         # Periodically poll if the snyk-monitor has upgraded
         for (( attempt=1; attempt<ATTEMPTS; attempt++))
         do
-          # Grab the tag of the snyk-monitor container image.
-          SNYK_MONITOR_POD=$(kubectl get pods -n snyk-monitor --no-headers | grep "snyk-monitor")
-          if [[ ! -z "${SNYK_MONITOR_POD}" ]]; then
```
minsiyang, line 184: So we don't need to check `SNYK_MONITOR_POD` might be empty then?

.circleci/config/jobs/operator_upgrade_tests.yml
```diff
@@ -175,21 +175,18 @@ steps:
...
-            VERSION=$(echo "${SNYK_MONITOR_POD}" | \
-            awk 'END { print $1 }' | \
+          # Grab the tag of the snyk-monitor container image. If snyk-monitor is not deployed for some reason, we exit immediately.
+          VERSION=$(kubectl get pods -n snyk-monitor --no-headers | \
+            grep "snyk-monitor" | \
+            awk 'END { if (NR==0) exit 1; else print $1 }' | \
```
minsiyang, line 184: Is it because if `VERSION` is empty, then NR would be 0?

.circleci/config/jobs/operator_upgrade_tests.yml
kat1906: I just reverted to the old way as it used to work 🤷

minsiyang (approved):

PR #1083, opened 2022-05-23
chore: fix skopeo install

yaronschwimmer:

### What this does
Skopeo was removed from the `kubic` repo, but is available from the official Ubuntu repo (from 20.10 and up). This caused the system checks to fail in CircleCI.
Upgrading the base image to Ubuntu 22.04 to fix skopeo install.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -952,7 +952,7 @@ jobs:
     system_tests:
         machine:
             docker_layer_caching: true
-            image: ubuntu-2004:202111-01
+            image: ubuntu-2204:current
         steps:
             - checkout
             - setup_node16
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -952,19 +952,14 @@ jobs:
     system_tests:
         machine:
             docker_layer_caching: true
-            image: ubuntu-2004:202111-01
+            image: ubuntu-2204:current
```
kat1906, line 955: is `current` ok to use?

yaronschwimmer: From what I understand (CircleCI have a terrible documentation), this tag is for the most recent image (with security patches etc.), so I think this should be fine

yaronschwimmer: For testing purposes, that is. I wouldn't do it for releases

yaronschwimmer: https://discuss.circleci.com/t/new-image-tag-convention-for-ubuntu-linux-android-and-windows-vms/43750

dkontorovskyy: https://discuss.circleci.com/t/ubuntu-linux-vm-machine-images-2022-april-q2-update/43749

PR #1086, opened 2022-05-25
fix: give openshift namespace watch the corrent permission

minsiyang:

### What this does

For openshift namespace watch, we didn't give it the same permission as the cluster watch. This PR tries to give it the same permission, and add logging for it.

### Notes for the reviewer

[slack thread](https://snyk.slack.com/archives/CDSMEJ29E/p1653398616088809)


src/supervisor/watchers/index.ts
```diff
@@ -60,9 +60,18 @@ export async function beginWatchingWorkloads(): Promise<void> {
...
+        config.WATCH_NAMESPACE,
+      );
+    } catch (err) {
+      logger.error(
+        { watchedNamespace: config.WATCH_NAMESPACE, err },
+        'failed to read the namespace',
+      );
+      return;
+    }
     const namespace = namespaceResponse.body;
```
kat1906, line 75: this will throw error here if `namespaceResponse.body` is not assigned in L65. Did you type it as `any` to get around the linter? 😆

src/supervisor/watchers/index.ts
```diff
@@ -60,9 +60,18 @@ export async function beginWatchingWorkloads(): Promise<void> {
       { namespace: config.WATCH_NAMESPACE },
       'kubernetes-monitor restricted to specific namespace',
     );
-    const namespaceResponse = await k8sApi.coreClient.readNamespace(
-      config.WATCH_NAMESPACE,
-    );
+    let namespaceResponse: any;
```
kat1906, line 63: why is this typed as any?

src/supervisor/watchers/index.ts
kat1906, line 75: I think we can't assume the key `body` will exist?

src/supervisor/watchers/index.ts
minsiyang: because it can be either the correct response, or it can be empty. I have changed it now.

src/supervisor/watchers/index.ts
kat1906: ie when `undefined`

minsiyang: if there is an error, the `namespaceResponse` would be undefined, otherwise, It would have a `body`

src/supervisor/watchers/index.ts
```diff
@@ -60,9 +60,18 @@ export async function beginWatchingWorkloads(): Promise<void> {
...
+    try {
+      namespaceResponse = await k8sApi.coreClient.readNamespace(
+        config.WATCH_NAMESPACE,
+      );
+    } catch (err) {
+      logger.error(
+        { watchedNamespace: config.WATCH_NAMESPACE, err },
+        'failed to read the namespace',
+      );
+      return;
```
ivanstanev, line 73: Just to note that this will just end the application so it will keep restarting in a loop 😄

kat1906: I think this is ok for the for running in 1 namespace? should we retry?

minsiyang: but now that we give the permission, it shouldn't return err. If they did return err, we want to log it and crash the app, so they will come to us with this error quicker?

PR #1174, opened 2022-10-14
fix: return pod image metadata for unsupported workload

minsiyang:

### What this does

return pod image metadata for unsupported workload kind Node

### DIFF
--- a/src/scanner/images/skopeo.ts
+++ b/src/scanner/images/skopeo.ts
@@ -36,6 +36,7 @@ export async function pull(
   image: string,
   destination: string,
   skopeoRepoType: SkopeoRepositoryType,
+  workloadName: string,
 ): Promise<void> {
   const creds = await credentials.getSourceCredentials(image);
   const credentialsParameters = getCredentialParameters(creds);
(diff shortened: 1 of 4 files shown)


src/scanner/images/skopeo.ts
```diff
@@ -77,7 +79,7 @@ async function pullWithRetry(
         }
       } catch (deleteErr) {
         logger.warn(
-          { error: deleteErr, destination },
+          { workloadName, error: deleteErr, destination },
```
ivanstanev, line 82: Nice 😄

src/supervisor/metadata-extractor.ts
```diff
@@ -215,9 +217,20 @@ export async function buildMetadataForWorkload(
   if (podOwner === undefined) {
```
ivanstanev, line 217: I wonder if to make this change a bit safer, we could also check if the pod.ownerReferences contain kind: Node? This way we fix this issue only for the specific static Pods case, without risking potential other bugs when the Pod has a regular workload as a parent but we couldn't retrieve it

PR #1183, opened 2022-10-21
test: jest fail-fast when any test fails

ivanstanev:

### What this does

Since Jest doesn't stop as soon as any test fails, we sometimes see weird errors and alerts because of an incorrectly configured test setup sending bad data.

This change ensures Jest fails fast on first error encountered.

### DIFF
--- a/test/integration/kubernetes.spec.ts
+++ b/test/integration/kubernetes.spec.ts
@@ -29,8 +29,9 @@ async function teardown(): Promise<void> {
   console.log('Removed "kind" network');
 }
 
-beforeAll(teardown);
 afterAll(teardown);
+
+test('clean up environment on start', teardown);
 
 // Make sure this runs first -- deploying the monitor for the next tests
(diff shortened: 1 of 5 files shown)


test/integration/kubernetes.spec.ts
```diff
@@ -29,9 +29,10 @@ async function teardown(): Promise<void> {
   console.log('Removed "kind" network');
 }
 
-beforeAll(teardown);
 afterAll(teardown);
 
+test('clean up environment on start', teardown);
```
kat1906, line 34: the beforeAll doesnt work?

kat1906 (approved):

test/integration/kubernetes.spec.ts
ivanstanev: Yes, seems like this setup-after-env only works for test() cases. The problem we had was in the beforeAll() where setup could fail but tests still continued. I tried to throw in the beforeAll() and it still continued doing stuff!

PR #1206, opened 2022-11-17
Fix/caching

ivanstanev:

### What this does

Various fixes for tracking workloads and caching them.


src/state.ts
```diff
@@ -43,9 +43,9 @@ function getWorkloadAlreadyScannedKey(
 
 function getWorkloadImageAlreadyScannedKey(
   workload: WorkloadAlreadyScanned,
-  imageId: string,
+  imageName: string,
 ): string {
-  return `${workload.namespace}/${workload.type}/${workload.uid}/${imageId}`;
+  return `${workload.namespace}/${workload.type}/${workload.uid}/${imageName}`;
```
kat1906, line 48: how about just `${workload.uid}/${imageName}`?

src/state.ts
```diff
@@ -72,19 +72,30 @@ export async function deleteWorkloadAlreadyScanned(
 
 export async function getWorkloadImageAlreadyScanned(
   workload: WorkloadAlreadyScanned,
+  imageName: string,
   imageId: string,
 ): Promise<string | undefined> {
-  const key = getWorkloadImageAlreadyScannedKey(workload, imageId);
-  return state.imagesAlreadyScanned.get(key);
+  const key = getWorkloadImageAlreadyScannedKey(workload, imageName);
+  const hasImageId = state.imagesAlreadyScanned.get(key)?.has(imageId);
```
kat1906, line 79: I think this will still cause the race condition? what happens if `sha:xxx-old-image` exists in the cache then `sha:yyy-new-image` comes along and replaces it but then another workload with `sha:xxx-old-image` comes along and overwrites it again?

kat1906: maybe we should keep a record of which shas we have seen per image name and just scan and send if it has not been seen before?

minsiyang: I think the `imagesLruCacheOptions` becoming a `Set` should prevent the overwriting, only the not existing image Id would be able to add to images (in L93)?

src/state.ts
minsiyang: 💡

src/state.ts
ivanstanev: >but then another workload

That gets a new key in the cache, so it won't overwrite the image 🙏 And now that we have a Set<>, we can track all occurrences of a name:tag+image-sha for a workload

kat1906: aha great! sorry, I misunderstood! 🤩

kat1906 (approved):
👏

PR #1210, opened 2022-11-28
fix: allow init containers to be configured

minsiyang:

### What this does

initContainers required to be run as root to provision pvc for temporary storage. We want to allow it to be configured, so user has the option to choose on installation.

### DIFF
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -148,6 +148,9 @@ sysdig:
 strategy:
   type: RollingUpdate
 
+initContainers:
+  enabled: true
+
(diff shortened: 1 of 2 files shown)


snyk-monitor/values.yaml
```diff
@@ -148,6 +148,9 @@ sysdig:
 strategy:
   type: RollingUpdate
 
+initContainers:
+  enabled: false
```
ivanstanev, line 152: It needs to be on by default, otherwise, it's a breaking change!

minsiyang: Oops 🙈

ivanstanev (approved):

PR #1221, opened 2022-12-07
RUN-2459 feat: pass image id to snyk docker plugin 

ivanstanev:

### What this does

Pass workload metadata image Id when available to Snyk Docker Plugin to collect image names as a new fact.

### DIFF
--- a/src/scanner/images/index.ts
+++ b/src/scanner/images/index.ts
@@ -123,6 +123,7 @@ export async function scanImages(
       const pluginResponse = await scan({
         path: archivePath,
         imageNameAndTag: imageName,
+        imageNameAndDigest: imageWithDigest,
       });
 
       if (
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "child-process-promise": "^2.2.1",
         "fs-extra": "^10.1.0",
         "lru-cache": "^6.0.0",
-        "minipass": "3.3.6",
         "needle": "^3.0.0",
         "sleep-promise": "^9.1.0",
         "snyk-config": "5.1.0",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "child-process-promise": "^2.2.1",
     "fs-extra": "^10.1.0",
     "lru-cache": "^6.0.0",
-    "minipass": "3.3.6",
     "needle": "^3.0.0",
     "sleep-promise": "^9.1.0",
     "snyk-config": "5.1.0",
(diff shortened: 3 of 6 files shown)


src/scanner/images/index.ts
```diff
@@ -119,10 +119,14 @@ export async function scanImages(
   for (const { imageName, fileSystemPath, imageWithDigest } of images) {
     try {
       const archivePath = `docker-archive:${fileSystemPath}`;
+      const imageNamesOverride: string[] = !!imageWithDigest
```
minsiyang, line 122: `imageWithDigest` should be `string`? Why does it need `!!`?

kat1906: `!!` turns it into a boolean to use in the ternary condition so if there is `imageWithDigest`, add it to the array with `imageName`, if not then the array contains just `imageName`

minsiyang: but if `imageWithDigest` is a valid string, on it is own in the ternary would return truthy value anyway, if the `imageWithDigest` is an empty string, then it would be falsy anyway, right?

src/scanner/images/index.ts
```diff
@@ -119,10 +119,13 @@ export async function scanImages(
   for (const { imageName, fileSystemPath, imageWithDigest } of images) {
     try {
       const archivePath = `docker-archive:${fileSystemPath}`;
+      const imageNameAndDigest: string | undefined =
+        imageWithDigest ?? undefined;
```
ivanstanev, line 123: I think the `??` operator will preserve the empty string, which is probably not what we want. Perhaps we could use `imageWithDigest | undefined`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
> ... is a logical operator that returns its right-hand side operand when its left-hand side operand is [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null) or [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined)

minsiyang: Oops, I missed that. I reversed the order, so now if `''`, `undefined` would be returned first.

src/scanner/images/index.ts
```diff
@@ -119,10 +119,13 @@ export async function scanImages(
   for (const { imageName, fileSystemPath, imageWithDigest } of images) {
     try {
       const archivePath = `docker-archive:${fileSystemPath}`;
+      const imageNameAndDigest: string | undefined =
+        undefined ?? imageWithDigest;
```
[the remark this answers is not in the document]
minsiyang: According to the logic here https://github.com/snyk/kubernetes-monitor/blob/staging/src/scanner/index.ts#L94-L108, imageWithDigest would always exist in the form of either `undefined` or `name@digest`. We can send this info to Docker Plugin straight away.

PR #1244, opened 2023-01-11
chore: upgrade deprecated k8s api for fixture

minsiyang:

### What this does

The extensions/v1beta1 would be deprecated after k8s v1.22
 https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122 
We want to update this to prevent having flaky tests in the future.

### DIFF
--- a/test/fixtures/sidecar-containers/deployment.yaml
+++ b/test/fixtures/sidecar-containers/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   annotations:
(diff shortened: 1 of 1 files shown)


test/fixtures/sidecar-containers/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
```
ivanstanev, line 1: Given the `kind: Deployment`, shouldn't apiVersion be apps/v1? https://github.com/snyk/kubernetes-monitor/blob/staging/test/fixtures/java-deployment.yaml#L1-L2

minsiyang: Probably https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment

Why is the `extensions/v1beta1` here first? Is it part of Deployment kind too?

ivanstanev (approved):

PR #1248, opened 2023-01-26
[CIT-75] feat: call authenticated snyk API endpoints with token

minsiyang:

### What this does

BREAKING CHANGE: Add service account token in Snyk Monitor secret to allow Snyk backend to validate user access when making API requests.


.circleci/config.yml
```diff
@@ -519,7 +519,8 @@ jobs:
 
                     set -e
                     INTEGRATION_ID=$(uuidgen)
-                    kubectl create secret generic snyk-monitor -n snyk-monitor --from-literal=integrationId=${INTEGRATION_ID} --from-literal=dockercfg.json={}
+                    SERVICE_ACCOUNT_API_TOKEN=$(uuidgen)
```
kat1906, line 522: this is not uuid and also will need to be valid in order for the requests to be accepted by k-u - not sure what the implications here are

src/transmitter/index.ts
```diff
@@ -65,7 +72,9 @@ export async function sendDepGraph(
       const request: KubernetesUpstreamRequest = {
         method: 'post',
         url: `${upstreamUrl}/api/v1/dependency-graph`,
+        // TODO: replace the url when Cerberus is configured for K-U
```
kat1906, line 75: 👍

.circleci/config.yml
snaftaly: Here we are only creating the SERVICE_ACCOUNT_API_TOKEN for starting the kubernetes monitor, but not really doing anything with it, therefore we can use a random uuid.

src/transmitter/index.ts
```diff
@@ -69,8 +92,10 @@ export async function sendDepGraph(
     try {
       const request: KubernetesUpstreamRequest = {
         method: 'post',
-        url: `${upstreamUrl}/api/v1/dependency-graph`,
+        url: constructUpstreamRequestUrl('/api/v1/dependency-graph'),
+        // TODO: replace the url when Cerberus is configured for K-U
```
iuliams, line 96: can the TODOs be removed?

iuliams (approved):

snyk-monitor/README.md
```diff
@@ -25,13 +25,21 @@ abcd1234-abcd-1234-abcd-1234abcd1234
 ```
 The Snyk Integration ID is used in the `--from-literal=integrationId=` parameter in the next step.
 
-2. (Optional) If you are not using any private registries, create a Kubernetes secret called `snyk-monitor` containing the Snyk Integration ID from the previous step running the following command:
+2. Create a Group or Org Service Account Token as described in [Snyk Service Account public documentation](https://docs.snyk.io/user-and-group-management/structure-account-for-high-application-performance/service-accounts). Make sure to use the role `Group Admin` or `Org Admin` accordingly.
```
kat1906, line 28: need to adjust for new permission

PR #1253, opened 2023-02-06
Fix/namespaced watch

ivanstanev:

### What this does

- update watch endpoints to non-deprecated APIs
- enqueue Kubernetes client API requests to prevent high load
- cache workload details to reduce chances of rate limiting
- use fastq as queueing library for API calls


src/state.ts
```diff
@@ -23,6 +28,28 @@ const workloadsLruCacheOptions: LruCache.Options<string, string> = {
   updateAgeOnGet: false,
 };
 
+/**
+ * Type is identical to the one in src/supervisor. However, it is declared here
+ * to prevent a dependency of this module to outside users and avoid circular refs.
+ * Additionally, we may want to know how our cache state is affected whenever the type changes.
+ */
+type IKubeObjectMetadataWithoutPodSpec = {
```
ahmed-agabani-snyk, line 36: `IKubeObjectMetadataWithoutPodSpec` can be moved from `src/state.ts` and `src/supervisor/workload-reader.ts` into `src/supervisor/types.ts` to resolve the circular reference.

ahmed-agabani-snyk (approved):

test/unit/supervisor/pod-watch-handler-caches.spec.ts
```diff
@@ -57,19 +57,19 @@ describe('image and workload image cache', () => {
   it('pushed data to send', async () => {
     await pod.podWatchHandler(podObject);
     await sleep(500);
-    expect(pushCallCount).toEqual(2);
+    expect(pushCallCount).toEqual(1);
```
minsiyang, line 60: Nit: Should we update the import on L1 and the mock on L7?

src/state.ts
```diff
@@ -137,6 +164,31 @@ export function deleteNamespace(namespace: V1Namespace): void {
...
 }
 
+function getWorkloadMetadataCacheKey(
+  workloadName: string,
+  namespace: string,
+): string {
+  return `${namespace}/${workloadName}`;
+}
+
+export function getWorkloadMetadata(
```
minsiyang, line 174: Nit: Could we rename the function to `getCachedWorkloadMetadata`, so we don't have to rename it in `state.ts`?

src/state.ts
```diff
@@ -137,6 +164,31 @@ export function deleteNamespace(namespace: V1Namespace): void {
...
+  workloadName: string,
+  namespace: string,
+): IKubeObjectMetadataWithoutPodSpec | undefined {
+  const key = getWorkloadMetadataCacheKey(workloadName, namespace);
+  const cachedMetadata = state.workloadMetadata.get(key);
+  return cachedMetadata;
+}
+
+export function setWorkloadMetadata(
```
minsiyang, line 183: Nit: similar to the above comment. 😄

minsiyang (approved):

src/state.ts
ivanstanev: There's no real circular dep per se, I just found it weird that state depends on the supervisor module, which depends on the state module 🤔

Issue #1277, opened 2023-03-02
[🐛]Cannot upgrade Helm Chart from "1.103.0" -> "2.0.0"


ajchiarello:
- `kubernetes-monitor version 1.103.0` 
- `Cloud runtime RKE2`

### Expected behaviour
The upgrade to succeed.

### Actual behaviour
The upgrade fails with this error:

`cannot patch "snyk" with kind Deployment: Deployment.apps "snyk" is invalid: [spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].values: Required value: must be specified when `operator` is 'In' or 'NotIn', spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[1].values: Required value: must be specified when `operator` is 'In' or 'NotIn']
`
### Steps to reproduce
Perform a helm update from 1.103.0 to 2.0.0. I am using the terraform helm provider, with the following values.yaml:
```
clusterName: "${var.project_name}"

excludedNamespaces:
${join("\n", formatlist("  - %s", local.exclude_namespaces))}

requests:
  cpu: "1500m"
  ephemeral-storage: "50Gi"
  memory: "2350Mi"

limits:
  cpu: "3200m"
  ephemeral-storage: "50Gi"
  memory: "5Gi"

strategy:
  type: Recreate
```

mihaibuzgau:
Thanks, @ajchiarello, for rasing this issue!
I've implemented a fix in this PR: this issue is fixed in: https://github.com/snyk/kubernetes-monitor/pull/1290

PR #1290, opened 2023-03-25
fix: helm chart when kubernetesIoArch value is missing

mihaibuzgau:
The current Kubernetes helm chart assumes that the value of kubernetesIoArch is set. If the parameter is not set, the creation of the helm chart fails with `Required value: must be specified when operatoris 'In' or 'NotIn'`

### What this does

Fixes https://github.com/snyk/kubernetes-monitor/issues/1277

### Notes for the reviewer

To reproduce the issue, you need to delete the following lines: https://github.com/snyk/kubernetes-monitor/blob/staging/snyk-monitor/values.yaml#L105-L106

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -45,15 +45,23 @@ spec:
                 - key: "kubernetes.io/arch"
                   operator: In
                   values:
-                    {{- with .Values.nodeAffinity.kubernetesIoArch }}
-                      {{- toYaml . | nindent 20 }}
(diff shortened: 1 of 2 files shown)


jonnyowenpowell (commented):
It seems surprising to have a default arch affinity - is there a reason we can't remove the matcher all together if the value isn't set?

jonnyowenpowell (approved):
LGTM 👍 

I do hold the opinion that as we don't own the nodes this will be deployed on, a default arch is essentially just a guess - but it will be right in a lot of cases, so I can't see it being a big problem.

PR #1326, opened 2023-06-22
fix: add envvar AZURE_CLIENT_ID option

popas90:
Add envvar AZURE_CLIENT_ID that allows specifying a managed identity to be used when pulling images for scanning.

### What this does

Updated Helm chart to add AZURE_CLIENT_ID envvar to `snyk-monitor` container, when needed (not empty string). The envvar is needed to avoid conflicting managed identities that allow the `snyk-monitor` to pull images from the container registry, therefore explicitly saying which of the identities to be used (by specifying its Client ID). 

### Notes for the reviewer

Will update documentation about setting up `kubernetes-monitor` in Azure.

### Screenshots

N/A

### DIFF
--- a/snyk-monitor/README.md
+++ b/snyk-monitor/README.md
@@ -308,6 +308,25 @@ helm upgrade --install snyk-monitor snyk-charts/snyk-monitor \
   --set volumes.projected.serviceAccountToken=true
 ```
 
+## Using AKS with Managed Identities
+
+For the particular case when you are using AKS with user-managed identities to authorize access to ACR and there are multiple identities that assign the `AcrPull` role to the VM scale set, you must also specify the Client ID of the desired user-managed identity to be used. This value must be set as an override, in `.Values.azureEnvVars`:
+```yaml
+azureEnvVars:
+  - name: AZURE_CLIENT_ID
+    value: "abcd1234-abcd-1234-abcd-1234abcd1234"
+```
+
+With the YAML above saved in `override.yaml`, run the following:
+
+```shell
+helm upgrade --install snyk-monitor snyk-charts/snyk-monitor \
+  --namespace snyk-monitor \
+  -f override.yaml
+```
+
+By default, this value is an empty string, and it will not be used as such.
+
 ## Configuring resources
 
 If more resources is required in order to deploy snyk-monitor, you can configure the helm charts default value for requests and limits with the `--set` flag.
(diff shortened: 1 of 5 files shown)


kat1906 (approved):

minsiyang (commented):
Can you also update the docs here too with how to set this up and why 🤗

snyk-monitor/templates/deployment.yaml
```diff
@@ -201,9 +201,15 @@ spec:
                 key: token
                 optional: true
           {{- end }}
-          {{- with .Values.envs }}
+          {{- with .Values.nodeEnvVars }}
           {{- toYaml . | trim | nindent 10 -}}
           {{- end }}
+          {{- range $v := .Values.azureEnvVars }}
```
minsiyang, line 207: 🤩

snyk-monitor/README.md
```diff
@@ -308,6 +308,17 @@ helm upgrade --install snyk-monitor snyk-charts/snyk-monitor \
   --set volumes.projected.serviceAccountToken=true
 ```
 
+## Using AKS with Managed Identities
+
+If you are using AKS with Managed Identities to authorize access to ACR, you must also specify the Client ID of the user-managed identity that assigns the `AcrPull` role to the VM scale set of that cluster. This value must be set as an override, in `.Values.azureEnvVars`:
```
minsiyang, line 313: If the user uses a `system-assigned` identity, then they won't need this step, and possibly just 1 `user-assigned` identity (not sure about this one 🤔). 
This env var would only need to be populated when the set up of the Managed Identity has multiple user-assigned managed identities. Maybe we should be explicit here to avoid further questions from users 😅

minsiyang (approved):

snyk-monitor/README.md
```diff
@@ -308,6 +308,17 @@ helm upgrade --install snyk-monitor snyk-charts/snyk-monitor \
...
+```yaml
+azureEnvVars:
+  - name: AZURE_CLIENT_ID
+    value: "abcd1234-abcd-1234-abcd-1234abcd1234"
+```
+
+By default, this value is an empty string, and it will not be used as such.
```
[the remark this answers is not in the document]
popas90: done!

ivanstanev (approved):

minsiyang (commented):
Nit: you can squash all 6 commits into one commit, and give it a meaningful prefix before the commit message 
https://www.notion.so/snyk/NOT-ARCHIVE-Git-Work-Methodologies-be6d298d764d47ca998d8f84991f4344?pvs=4#f2af5841f22446f28b3c27aec7a53642
eg, `fix: add envvar AZURE_CLIENT_ID option`

PR #1353, opened 2023-08-21
fix: UBI9 image build

jonnyowenpowell:

### What this does

This PR improves the installation and verification of dependencies in the UBI9 image build.

Dependency installation via package managers or builds from source have been replaced by binary downloads. Binaries are downloaded with `curl`s `-f (--fail)` set, and hashes of the downloaded files are checked before continuing the build. A notable exception is `gcloud`, the Google Cloud CLI, which continues to be installed via their provided script. Python 3 is no longer explicitly installed, and instead we allow `gcloud` to install its bundled version of Python.

The application is now built using the `ubi8/nodejs-16` base image, to align more closely with the production container's base image. Due to the age of Node.js 16, a UBI9 base image is not available. When this project updates to use Node.js 18, we can build using the `ubi9/nodejs-18` image, achieving complete alignment with the production container's base image.

Platforms are explicitly set to `linux/amd64` for both build stages, to ensure that developer's machine architecture does not affect the image being built.

### Notes for the reviewer

This resolves a couple of issues:
- Skopeo failing to find the shared library `libdevmapper`
- Failed downloads of binaries resulting in failure messages from the remote being naively saved as executable binaries

### DIFF
--- a/scripts/docker/build-image-ubi9.sh
+++ b/scripts/docker/build-image-ubi9.sh
@@ -1,4 +1,7 @@
 #! /bin/bash
+set -e
+
+DIR="$(cd "$(dirname "$0")" && pwd)"
 
 # default value for our name and tag, when we don't want to push the image
 # for example when testing locally or on opening a PR
@@ -19,3 +22,5 @@ docker build \
   --build-arg NODE_16_LATEST_VERSION_TAR_GZ_FILE_SHASUM256="${NODE_16_LATEST_VERSION_TAR_GZ_FILE_SHASUM256}" \
   -t ${NAME_AND_TAG} \
   --file=Dockerfile.ubi9 .
+
+"$DIR/smoke-test-image-binaries.sh" $NAME_AND_TAG
--- a/Dockerfile.ubi9
+++ b/Dockerfile.ubi9
@@ -1,46 +1,50 @@
-#---------------------------------------------------------------------
-# STAGE 1: Build skopeo inside a temporary container
-#---------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


scripts/docker/build-image-ubi9.sh
```diff
@@ -19,3 +20,8 @@ docker build \
   --build-arg NODE_16_LATEST_VERSION_TAR_GZ_FILE_SHASUM256="${NODE_16_LATEST_VERSION_TAR_GZ_FILE_SHASUM256}" \
   -t ${NAME_AND_TAG} \
   --file=Dockerfile.ubi9 .
+
+# Smoke test image binaries
+docker run --entrypoint /usr/bin/dumb-init ${NAME_AND_TAG} --version
```
iuliams, line 25: Nice! What about adding some extra flags to ensure:
* it doesn't hang for a long time (e.g, `docker run [...] timeout 20`)
* the container is removed upon exit (e.g `docker run --rm [...]`)

jonnyowenpowell: `dumb-init` is in control so the timeout won't be able to function anyway if it's broken, and if it's working then it'll exit quick. It would be nice to have some way of stopping it hanging though 🤔 `--rm` doesn't matter in CI but is good locally so I'll add it.

jonnyowenpowell: I've added a bash timeout to the whole `docker run` command.

iuliams (approved):

PR #1362, opened 2023-09-05
chore: use "fix" commit type when upgrading dependencies

gergo-papp:
### What this does

See internal slack discussion: https://snyk.slack.com/archives/C07N668DA/p1693564826434639


kat1906:
this workflow never passes, might be better to just remove it from this repo

iuliams (approved):

PR #1369, opened 2023-09-14
chore: Update and relock NPM packages

snyk-deployer:
This PR updates and relocks NPM packages.

It is created by a recurring github action checking for outdated dependencies.


iuliams:
PRs with newer versions were created.

PR #1397, opened 2023-10-30
Fix/package json

iuliams:
Attempt to fix the conflict between staging and master.


iuliams:
Wrong branches picked.

PR #1419, opened 2023-12-01
feat: use Sysdig EVE public API [CIT-704]

iuliams:

### What this does
Migrates to Sysdig v2 EVE API.

Screenshot 2023-12-13 at 10 29 45

### DIFF
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -188,17 +188,23 @@ spec:
           - name: SNYK_WORKERS_COUNT
             value: {{ quote .Values.workers.count }}
           {{- if .Values.sysdig.enabled }}
-          - name: SNYK_SYSDIG_ENDPOINT
+          - name: SNYK_SYSDIG_REGION_URL
             valueFrom:
               secretKeyRef:
                 name: {{ .Values.sysdig.secretName }}
-                key: endpoint
+                key: region
                 optional: true
-          - name: SNYK_SYSDIG_TOKEN
+          - name: SNYK_SYSDIG_API_TOKEN
             valueFrom:
               secretKeyRef:
                 name: {{ .Values.sysdig.secretName }}
                 key: token
+                optional: true
+          - name: SNYK_SYSDIG_CLUSTER_NAME
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.sysdig.secretName }}
+                key: cluster
                 optional: true
           {{- end }}
           {{- with .Values.envs }}
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -188,23 +188,35 @@ spec:
           - name: SNYK_WORKERS_COUNT
             value: {{ quote .Values.workers.count }}
           {{- if .Values.sysdig.enabled }}
+          - name: SNYK_SYSDIG_RISK_SPOTLIGHT_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.sysdig.secretName }}
+                key: token
+                optional: true
           - name: SNYK_SYSDIG_REGION_URL
             valueFrom:
               secretKeyRef:
                 name: {{ .Values.sysdig.secretName }}
                 key: region
                 optional: true
-          - name: SNYK_SYSDIG_API_TOKEN
-            valueFrom:
-              secretKeyRef:
-                name: {{ .Values.sysdig.secretName }}
-                key: token
-                optional: true
           - name: SNYK_SYSDIG_CLUSTER_NAME
             valueFrom:
               secretKeyRef:
                 name: {{ .Values.sysdig.secretName }}
                 key: cluster
+                optional: true
+          - name: SNYK_SYSDIG_ENDPOINT
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.sysdig.legacySecretName }}
+                key: endpoint
+                optional: true
+          - name: SNYK_SYSDIG_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.sysdig.legacySecretName }}
+                key: token
                 optional: true
           {{- end }}
           {{- with .Values.envs }}
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -209,13 +209,13 @@ spec:
           - name: SNYK_SYSDIG_ENDPOINT
             valueFrom:
               secretKeyRef:
-                name: {{ .Values.sysdig.legacySecretName }}
+                name: {{ .Values.sysdig.secretName }}
                 key: endpoint
                 optional: true
           - name: SNYK_SYSDIG_TOKEN
             valueFrom:
               secretKeyRef:
-                name: {{ .Values.sysdig.legacySecretName }}
+                name: {{ .Values.sysdig.secretName }}
                 key: token
                 optional: true
           {{- end }}
--- a/snyk-monitor/templates/deployment.yaml
+++ b/snyk-monitor/templates/deployment.yaml
@@ -191,19 +191,19 @@ spec:
           - name: SNYK_SYSDIG_RISK_SPOTLIGHT_TOKEN
             valueFrom:
               secretKeyRef:
-                name: {{ .Values.sysdig.secretName }}
+                name: snyk-sysdig-secret
                 key: token
                 optional: true
           - name: SNYK_SYSDIG_REGION_URL
             valueFrom:
               secretKeyRef:
-                name: {{ .Values.sysdig.secretName }}
+                name: snyk-sysdig-secret
                 key: region
                 optional: true
           - name: SNYK_SYSDIG_CLUSTER_NAME
             valueFrom:
               secretKeyRef:
-                name: {{ .Values.sysdig.secretName }}
+                name: snyk-sysdig-secret
                 key: cluster
                 optional: true
           - name: SNYK_SYSDIG_ENDPOINT
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -155,8 +155,7 @@ workers:
   count: 5
 sysdig:
   enabled: false
-  namespace: sysdig-agent
-  secretName: sysdig-eve-secret
+  secretName: snyk-sysdig-secret
 
 strategy:
   type: RollingUpdate
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -156,8 +156,7 @@ workers:
 
 sysdig:
   enabled: false
-  secretName: snyk-sysdig-secret
-  legacySecretName: sysdig-eve-secret
+  secretName: sysdig-eve-secret
 
 strategy:
   type: RollingUpdate
--- a/snyk-monitor/values.yaml
+++ b/snyk-monitor/values.yaml
@@ -153,9 +153,11 @@ skopeo:
 
 workers:
   count: 5
+
 sysdig:
   enabled: false
   secretName: snyk-sysdig-secret
+  legacySecretName: sysdig-eve-secret
 
 strategy:
   type: RollingUpdate
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -47,12 +47,6 @@ config.EXCLUDED_NAMESPACES = loadExcludedNamespaces();
 config.WORKERS_COUNT = Number(config.WORKERS_COUNT) || 10;
 config.SKOPEO_COMPRESSION_LEVEL = Number(config.SKOPEO_COMPRESSION_LEVEL) || 6;
 
-// return Sysdig v1 endpoint information
-if (config.SYSDIG_ENDPOINT && config.SYSDIG_TOKEN) {
-  config.SYSDIG_ENDPOINT = config.SYSDIG_ENDPOINT.trim();
-  config.SYSDIG_TOKEN = config.SYSDIG_TOKEN.trim();
-}
-
 // return Sysdig v2 endpoint information
 if (
   config.SYSDIG_RISK_SPOTLIGHT_TOKEN &&
@@ -63,6 +57,12 @@ if (
     config.SYSDIG_RISK_SPOTLIGHT_TOKEN.trim();
   config.SYSDIG_REGION_URL = config.SYSDIG_REGION_URL.trim();
   config.SYSDIG_CLUSTER_NAME = config.SYSDIG_CLUSTER_NAME.trim();
+}
+
+// return Sysdig v1 endpoint information
+if (config.SYSDIG_ENDPOINT && config.SYSDIG_TOKEN) {
+  config.SYSDIG_ENDPOINT = config.SYSDIG_ENDPOINT.trim();
+  config.SYSDIG_TOKEN = config.SYSDIG_TOKEN.trim();
 }
 
 /**
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -47,14 +47,21 @@ config.EXCLUDED_NAMESPACES = loadExcludedNamespaces();
 config.WORKERS_COUNT = Number(config.WORKERS_COUNT) || 10;
 config.SKOPEO_COMPRESSION_LEVEL = Number(config.SKOPEO_COMPRESSION_LEVEL) || 6;
 
-// return Sysdig endpoint information
+// return Sysdig v1 endpoint information
+if (config.SYSDIG_ENDPOINT && config.SYSDIG_TOKEN) {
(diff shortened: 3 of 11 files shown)


src/data-scraper/index.ts
```diff
@@ -50,7 +90,56 @@ export async function validateConnectivity(): Promise<void> {
...
+    },
+  };
+
+  while (true) {
+    try {
+      logger.info('attempting to get runtime images');
+
+      const { response, attempt } = await retryRequest(
+        'get',
+        `${url}?clusterName=${clusterName}`,
```
iuliams, line 111: I removed pagination as I had some issues with it. Investigating now.

snyk-monitor/values.yaml
```diff
@@ -158,6 +158,10 @@ sysdig:
   namespace: sysdig-agent
   secretName: sysdig-eve-secret
 
+sysdigV2:
```
iuliams, line 161: Would something else make it easier for the customers?
Will it make sense, once the migration is over, for customers to have to configure `sysdigV2` instead of `sysdig`?

iuliams: Option: Remove the need for this altogether and check if the secret was created.

snyk-monitor/values.yaml
```diff
@@ -158,6 +158,10 @@ sysdig:
...
+  enabled: false
+  secretName: snyk-sysdig-secret
```
iuliams, line 163: Review if this name shall be named differently.

src/healthcheck.ts
```diff
@@ -21,12 +24,26 @@ async function healthCheck(): Promise<void> {
...
 
 async function sysdigHealthCheck(): Promise<void> {
-  if (!config.SYSDIG_ENDPOINT || !config.SYSDIG_TOKEN) {
+  let sysdigVersion = '';
+  if (
+    config.SYSDIG_CLUSTER_NAME &&
+    config.SYSDIG_API_TOKEN &&
+    config.SYSDIG_REGION_URL
+  ) {
+    sysdigVersion = SYSDIG_V2;
```
kat1906, line 33: lets log the sysdig version too

snyk-monitor/values.yaml
```diff
@@ -46,7 +46,7 @@ clusterName: ""
 
 # The snyk-monitor requires disk storage to temporarily pull container images and to scan them for vulnerabilities.
 # This value controls how much disk storage _at most_ may be allocated for the snyk-monitor. Unless overridden by the `pvc` value, the snyk-monitor mounts an emptyDir for storage.
-temporaryStorageSize: 50Gi #  Applies to PVC too
+temporaryStorageSize: 30Gi #  Applies to PVC too
```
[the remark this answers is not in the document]
iuliams: Yeah... but just to match out cluster settings...
So, good catch, have to revert this change, since in our docs we require 50G.

snyk-monitor/templates/deployment.yaml
```diff
@@ -188,18 +188,24 @@ spec:
...
               secretKeyRef:
                 name: {{ .Values.sysdig.secretName }}
                 key: token
                 optional: true
+          - name: SNYK_SYSDIG_CLUSTER_NAME
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.sysdig.secretName }}
+                key: cluster
+                optional: true
```
kat1906, line 208: are they optional though?

src/data-scraper/index.ts
```diff
@@ -28,7 +36,8 @@ function isSuccessStatusCode(statusCode: number | undefined): boolean {
 /** NOTE: This function can throw, so the caller should handle errors. */
 export async function validateConnectivity(): Promise<void> {
   const url = getSysdigUrl();
-  const header = getSysdigAuthHeader();
+  const header = getSysdigAuthToken();
+  const clusterName = getSysdigClusterName();
```
kat1906, line 40: we can use `config.SYSDIG_CLUSTER_NAME` directly as it is not being modified

snyk-monitor/templates/deployment.yaml
iuliams: They should be there, unless a further check fails anyway, if they're not configure.
I kept it optional as it was for the previous secret... Any idea why it'd have been optional before?

src/data-scraper/index.ts
```diff
@@ -52,25 +59,23 @@ export async function validateConnectivity(): Promise<void> {
...
-  let cursor: string = '';
   while (true) {
     try {
-      logger.info({ cursor }, 'attempting to get runtime images');
+      logger.info('attempting to get runtime images');
 
       const { response, attempt } = await retryRequest(
         'get',
-        `${url}?limit=${limit}&cursor=${cursor}`,
+        `${url}?clusterName=${clusterName}`,
```
kat1906, line 78: is this API not paginated then?

iuliams: I'm getting a 500 when using pagination as before - so I'd say yes.

src/healthcheck.ts
```diff
@@ -21,7 +21,14 @@ async function healthCheck(): Promise<void> {
...
 async function sysdigHealthCheck(): Promise<void> {
-  if (!config.SYSDIG_ENDPOINT || !config.SYSDIG_TOKEN) {
+  if (
+    !(
+      config.SYSDIG_CLUSTER_NAME &&
+      config.SYSDIG_API_TOKEN &&
+      config.SYSDIG_REGION_URL
+    )
+  ) {
+    logger.info('Sysdig integration not configured');
```
kat1906, line 31: we don't need this log, it will trigger every time the healthcheck is called

snyk-monitor/templates/deployment.yaml
jonnyowenpowell: I don't think we should introduce a breaking change by making them non-optional. If some customers currently have `sysdig.enabled` set to `true` but haven't got around to actually adding the secrets for whatever reason, we break the entire monitor by making them non-optional.
I agree that it probably would've helped customers debug things easier if they weren't optional, but I don't think now is the time to change it.

kat1906: why not? this is a breaking change anyway

src/data-scraper/index.ts
kat1906: interesting, maybe ask on the channel for that, the payloads can be very large

kat1906: if we are not paginating we can remove the while(){} in L70

snyk-monitor/templates/deployment.yaml
jonnyowenpowell: to customers with a fully functioning sysdig integration yes, but not everyone

src/index.ts
```diff
@@ -68,11 +68,19 @@ async function monitor(): Promise<void> {
...
-  if (!config.SYSDIG_ENDPOINT || !config.SYSDIG_TOKEN) {
-    logger.info({}, 'Sysdig integration not detected');
+  if (
+    !(
+      config.SYSDIG_REGION_URL &&
+      config.SYSDIG_API_TOKEN &&
+      config.SYSDIG_CLUSTER_NAME
+    )
+  ) {
+    logger.info({}, 'Sysdig not properly configured');
```
kat1906, line 78: this will be logged for all customers not using the sysdig integration so we could do something like this:
```
  if (
    !config.SYSDIG_REGION_URL &&
    !config.SYSDIG_API_TOKEN &&
    !config.SYSDIG_CLUSTER_NAME
  ) {
    console.log('Sysdig integration not detected');
    return;
  }
  if (
    !config.SYSDIG_REGION_URL ||
    !config.SYSDIG_API_TOKEN ||
    !config.SYSDIG_CLUSTER_NAME
  ) {
    console.log('Sysdig integration not configured correctly');
    return;
  }
  console.log('Sysdig integration configured correctly');
  return;
```
to differentiate between the 3 states

alternatively, we can make the secret values not optional

snyk-monitor/templates/deployment.yaml
kat1906: do we really need to account for customers who have set the `sysdig.enabled` to `true` but are not actually using sysdig? seems a bit of a weird edge case

jonnyowenpowell: If even one exists, we might avoid a support ticket by keeping them optional!
My rationale here is that because this is a short term solution, we should limit the impact of this change as much as possible.

kat1906: we might avoid a support ticket by making them not optional 😆 anyways I don't mind

test/unit/data-scraper/scrape-data.spec.ts
```diff
@@ -32,33 +34,16 @@ describe('dataScraper()', () => {
...
+        '/api/scanning/eveintegration/v2/runtimeimages?clusterName=test-sysdig-cluster&limit=10',
+      )
       .times(1)
       .reply(200, bodyWithToken);
 
-    nock('https://sysdig', { reqheaders: { authorization: expectedHeader } })
-      .get('/v1/runtimeimages?limit=10&cursor=xxx')
-      .times(1)
-      .reply(200, bodyNoToken);
-
```
kat1906, line 61: why has this been removed? the test should follow the same patter ie should request the second page?

iuliams: Ah, just missed adding it back - i removed it when I initially removed pagination, because then it made no more sense. Thanks for noticing it.

kat1906 (approved):

snyk-monitor/values.yaml
```diff
@@ -153,10 +153,11 @@ skopeo:
 
 workers:
   count: 5
+
 sysdig:
   enabled: false
-  namespace: sysdig-agent
-  secretName: sysdig-eve-secret
+  secretName: snyk-sysdig-secret
+  legacySecretName: sysdig-eve-secret
```
kat1906, line 160: this might break a customer's deployment if they have renamed their secret

src/data-scraper/index.ts
```diff
@@ -29,19 +40,19 @@ function isSuccessStatusCode(statusCode: number | undefined): boolean {
 export async function validateConnectivity(): Promise<void> {
   const url = getSysdigUrl();
   const header = getSysdigAuthHeader();
+  const clusterName = config.SYSDIG_CLUSTER_NAME;
```
kat1906, line 43: this will not be set for legacy so will the url in L44 be correct?

src/data-scraper/index.ts
```diff
@@ -29,19 +40,19 @@ function isSuccessStatusCode(statusCode: number | undefined): boolean {
...
     },
     timeout: 10_000,
   };
-
   const limit: number = 1;
-  const cursor: string = '';
   const { response } = await retryRequest(
     'get',
-    `${url}?limit=${limit}&cursor=${cursor}`,
+    `${url}?clusterName=${clusterName}&limit=${limit}`,
```
kat1906, line 55: this url was `${url}?limit=${limit}&cursor=${cursor}` for legacy so will not work here?

kat1906: oh I see it below!

src/data-scraper/index.ts
kat1906: I see it below!

src/healthcheck.ts
```diff
@@ -20,13 +23,39 @@ async function healthCheck(): Promise<void> {
...
+  ) {
     return;
   }
 
   try {
-    await dataScraper.validateConnectivity();
+    let sysdigVersion = getSysdigVersion();
+    if (sysdigVersion == sysdigV1) {
+      await dataScraper.validateConnectivityV1();
+    } else if (sysdigVersion == sysdigV2) {
```
kat1906, line 56: I think this can just be `else` rather than `else if`

src/index.ts
```diff
@@ -10,9 +10,12 @@ import { beginWatchingWorkloads } from './supervisor/watchers';
 import { loadAndSendWorkloadEventsPolicy } from './common/policy';
 import { sendClusterMetadata } from './transmitter';
 import { setSnykMonitorAgentId } from './supervisor/agent';
-import { scrapeData } from './data-scraper';
+import { scrapeData, scrapeDataV1 } from './data-scraper';
 import { setupHealthCheck } from './healthcheck';
 
+export const sysdigV1 = 'V1';
+export const sysdigV2 = 'V2';
+
```
kat1906, line 18: these consts are also being exported from `src/healthcheck.ts`, I think we only need to export from one place

src/index.ts
```diff
@@ -67,16 +70,44 @@ async function monitor(): Promise<void> {
   }
 }
 
+export function getSysdigVersion() {
```
kat1906, line 73: same with this function!

src/index.ts
```diff
@@ -67,16 +70,44 @@ async function monitor(): Promise<void> {
...
+  let sysdigVersion = getSysdigVersion();
+  logger.info({}, `Sysdig ${sysdigVersion} data scraping starting`);
+
   const initialInterval: number = 20 * 60 * 1000; // 20 mins in milliseconds
   setTimeout(async () => {
     try {
-      await scrapeData();
+      if (sysdigVersion == sysdigV1) {
+        await scrapeDataV1();
+      } else if (sysdigVersion == sysdigV2) {
```
kat1906, line 108: same here: I think this can just be `else` rather than `else if`

src/index.ts
```diff
@@ -88,7 +119,11 @@ async function setupSysdigIntegration(): Promise<void> {
   const interval: number = 4 * 60 * 60 * 1000; // 4 hours in milliseconds
   setInterval(async () => {
     try {
-      await scrapeData();
+      if (sysdigVersion == sysdigV1) {
+        await scrapeDataV1();
+      } else if (sysdigVersion == sysdigV2) {
```
kat1906, line 124: same here: I think this can just be `else` rather than `else if`

src/index.ts
iuliams: ah, got some circular import issues, and then missed to unexport them - redoing.

kat1906 (commented):
I think we should either change the identity.type for the new one (e.g. `sysdigV2`) or add another property here for the version https://github.com/snyk/kubernetes-monitor/blob/5b824472a12103f44ddc5959f20978474186615e/src/transmitter/types.ts#L130C11-L130C26 

for both options it will require changes to the API but otherwise we have no way of tracking if the old version is. The first option would not require a new release of the API but the second option probably will

snyk-monitor/values.yaml
iuliams: Auch, good point. Thinking about an alternative.

src/common/config.ts
```diff
@@ -47,12 +47,24 @@ config.EXCLUDED_NAMESPACES = loadExcludedNamespaces();
 config.WORKERS_COUNT = Number(config.WORKERS_COUNT) || 10;
 config.SKOPEO_COMPRESSION_LEVEL = Number(config.SKOPEO_COMPRESSION_LEVEL) || 6;
 
-// return Sysdig endpoint information
+// return Sysdig v1 endpoint information
 if (config.SYSDIG_ENDPOINT && config.SYSDIG_TOKEN) {
```
kat1906, line 51: I am going to reorder these if statements incase the customer still has the old secret

snyk-monitor/templates/deployment.yaml
```diff
@@ -188,6 +188,24 @@ spec:
           - name: SNYK_WORKERS_COUNT
             value: {{ quote .Values.workers.count }}
           {{- if .Values.sysdig.enabled }}
+          - name: SNYK_SYSDIG_RISK_SPOTLIGHT_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: snyk-sysdig-secret
+                key: token
+                optional: true
```
jonnyowenpowell, line 196: Is there ever potential for another token? Should we call this one something more specific?

kat1906 (approved):

src/data-scraper/index.ts
```diff
@@ -13,12 +13,23 @@ const httpsAgent = new HttpsAgent({
   rejectUnauthorized: false,
 });
 
-function getSysdigUrl(): string {
+// Soon to be deprecated
+function getSysdigV1Url(): string {
```
jonnyowenpowell, line 17: nit: separating v1 into it's own file would make this easier to read, and make v1 easier to remove later.

jonnyowenpowell (approved):
LGTM 👍 Some minor points and nits, but would be happy enough if this went out as it 👍

snyk-monitor/templates/deployment.yaml
iuliams: I'd say it makes sense to be one token per secret, given that the secret hold info about 1 integration with the agent.

src/data-scraper/index.ts
iuliams: sure, will do that in a new PR.

PR #1422, opened 2023-12-13
chore: deprecate CronJob v1beta1

iuliams:

### What this does
Once this is deployed, we will remove support for  CronJob v1beta1 since it's been deprecated in the Kubernetes API.
This way, we allow for fast vulnerability fixes.


src/supervisor/watchers/handlers/cron-job.ts
```diff
@@ -137,7 +91,7 @@ export async function cronJobWatchHandler(
 async function isNamespacedCronJobSupportedWithClient(
```
kat1906, line 91: I don't think we need this function at all now, CronJobs should always be supported

src/supervisor/watchers/handlers/cron-job.ts
```diff
@@ -191,19 +147,12 @@ export async function isNamespacedCronJobSupported(
...
-  }
-  return await isNamespacedCronJobSupportedWithClient(
-    workloadKind,
-    namespace,
-    k8sApi.batchUnstableClient,
-  );
+  return false;
 }
 
 export async function isClusterCronJobSupportedWithClient(
```
kat1906, line 153: this can also be removed

src/supervisor/watchers/handlers/index.ts
```diff
@@ -23,11 +23,6 @@ async function isSupportedNamespacedWorkload(
       return await deploymentConfig.isNamespacedDeploymentConfigSupported(
         namespace,
       );
-    case WorkloadKind.CronJobV1Beta1:
-      return await cronJob.isNamespacedCronJobSupported(
-        workloadKind,
-        namespace,
-      );
     case WorkloadKind.CronJob:
```
kat1906, line 26: can also remove this case as the switch defaults to true (and cronjobs should always be supported now)

src/supervisor/workload-reader.ts
```diff
@@ -242,7 +242,7 @@ const jobReader: IWorkloadReaderFunc = async (workloadName, namespace) => {
   return metadata;
 };
 
-// cronJobReader can read v1 and v1beta1 CronJobs
+// cronJobReader can read v1 CronJobs
```
kat1906, line 245: I think this comment can be removed

src/supervisor/workload-reader.ts
```diff
@@ -253,13 +253,11 @@ const cronJobReader: IWorkloadReaderFunc = async (workloadName, namespace) => {
     .retryKubernetesApiRequest(() =>
       k8sApi.batchClient.readNamespacedCronJob(workloadName, namespace),
     )
-    // In case the V1 client fails, try using the V1Beta1 client.
+    // In case the V1 client fails, log it and throw an error.
```
kat1906, line 256: and this comment can be removed

src/supervisor/metadata-extractor.ts
```diff
@@ -161,7 +161,6 @@ export async function buildMetadataForWorkload(
       workloadName: pod.metadata?.name,
       workloadType: pod.kind,
       workloadNamespace: pod.metadata?.namespace,
-      clusterName: pod.metadata?.clusterName,
```
kat1906, line 164: why has this been removed?

iuliams: was removed from the pod's metadata.

src/supervisor/watchers/handlers/types.ts
```diff
@@ -101,6 +101,7 @@ export type V1ClusterList<T> = (
   pretty?: string,
   resourceVersion?: string,
   resourceVersionMatch?: string,
+  sendInitialEvents?: boolean,
```
kat1906, line 104: what is this for?

iuliams: Required by the latest Kubernetes API.
https://kubernetes.io/docs/reference/using-api/api-concepts/#streaming-lists

src/supervisor/workload-reader.ts
```diff
@@ -253,13 +252,10 @@ const cronJobReader: IWorkloadReaderFunc = async (workloadName, namespace) => {
...
-    // In case the V1 client fails, try using the V1Beta1 client.
     .catch(() =>
-      kubernetesApiWrappers.retryKubernetesApiRequest(() =>
-        k8sApi.batchUnstableClient.readNamespacedCronJob(
-          workloadName,
-          namespace,
-        ),
+      logger.info(
+        { workloadName, namespace },
+        'kubernetes api could not return workload',
```
kat1906, line 258: not sure on this log, in the jobReader above we do not catch the error, lets follow that pattern?

src/supervisor/watchers/handlers/types.ts
kat1906: what value do we set for this? or just undefined?

iuliams: It can be undefined and we do not set it -> so undefined.
e.g it's passed as a param in all the calls to `paginatedNamespacedList`

kat1906 (approved):

PR #1452, opened 2024-02-09
feat: produce oci distribution metadata [OI-61]

jonnyowenpowell:

### What this does

This PR updates `snyk-docker-plugin` and enhances the information passed to it's `scan` method to enable the `ociDistributionMetadata` fact to be included in Scan Results produced by this service.

### Notes for the reviewer


src/scanner/images/skopeo.ts
```diff
@@ -57,30 +78,38 @@ export async function pull(
...
   args: Array<processWrapper.IProcessArgument>,
+  env: Record<string, string | undefined>,
   destination: string,
   workloadName: string,
 ): Promise<void> {
   const MAX_ATTEMPTS = 10;
   const RETRY_INTERVAL_SEC = 0.2;
 
-  for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
+  let attempt = 1;
```
jonnyowenpowell, line 106: This refactor from `for` to `while` was to support returning something other than `void` here, but now we're not doing that, it might be worth putting it back how it was to avoid the chance of any accidental regressions here.
Passing `env` in still makes sense because we're sharing it with the other call now.

I do _like_ that we're not returning the digests from this function anymore, nice one!

src/scanner/images/skopeo.ts
```diff
@@ -1,12 +1,33 @@
...
+  mediaType: string;
+  manifests: Array<{
+    digest: string;
+    platform: {
+      architecture: string;
+      os: string;
+      variant: string;
+    };
+  }>;
+};
```
jonnyowenpowell, line 26: nit: these could go in the `types` file in this directory. They are 'private' though, in the sense that they're not needed elsewhere, so up to you.

jonnyowenpowell (commented):
This is looking really good! 👍
The only thing that IMO really needs fixing before this goes out is the `credentialsParameters` and `certificatesParameters` passing through to the skopeo inspect call.

ahmed-agabani-snyk (approved):

kat1906:
please can you amment the commit message to remove the `[wip]`?
[feat: pass oci distribution metadata to snyk-docker-plugin [wip]](https://github.com/snyk/kubernetes-monitor/pull/1452/commits/89c58cfe750f8ee6a04020a68da2976c27bafa4d)

PR #1465, opened 2024-03-04
fix: avoid sending empty runtime images payloads

jonnyowenpowell:
- (n/a) Documentation written [ℹ︎](https://github.com/snyk/general/wiki/Documentation)

### What this does

This PR modifies the Sysdig (v1 and v2) integration to avoid sending runtime data payloads upstream when all runtime images have been filtered out - usually due to being in an ignored namespace.

### DIFF
--- a/src/data-scraper/index.ts
+++ b/src/data-scraper/index.ts
@@ -101,8 +101,10 @@ export async function scrapeData(): Promise<void> {
         2,
       );
 
-      logger.info({}, 'sending runtime data upstream');
-      await sendRuntimeData(runtimeDataPayload);
+      if (runtimeDataPayload) {
+        logger.info({}, 'sending runtime data upstream');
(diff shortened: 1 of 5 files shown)


src/data-scraper/index.ts
```diff
@@ -101,8 +101,10 @@ export async function scrapeData(): Promise<void> {
         2,
       );
 
-      logger.info({}, 'sending runtime data upstream');
```
kat1906, line 104: should we have a log with something about no runtime results retrieved? It is good to be explicit in this component because it runs on a customers cluster

jonnyowenpowell: Well we did receive them, but we just filtered them all out - it's not surprising given that we only pull 10 runtime images at a time that sometimes we're not interested in any of them.

PR #1506, opened 2024-07-17
Fix/memory leak

crossleydominic:

### What this does

```
The crux of the issue is this:
➜  kubernetes-monitor git:(master) ✗ kubectl get --raw "/api/v1/watch/namespaces/services" | jq .
{
  // Irrelevant fields omitted, not the resourceVersion

  "type": "ADDED",
  "object": {
    "kind": "Namespace",
    "apiVersion": "v1",
    "metadata": {
      "resourceVersion": "616",
    },
}
```

```
➜  kubernetes-monitor git:(master) ✗ kubectl get --raw "/api/v1/watch/namespaces/services?resourceVersion=616" | jq .
{
  "type": "ERROR",
  "object": {
    "kind": "Status",
    "apiVersion": "v1",
    "metadata": {},
    "status": "Failure",
    "message": "The resourceVersion for the provided watch is too old.",
    "reason": "Expired",
    "code": 410
  }
}
```

version 616, is valid, it’s still 616 after making the second call, it’s not become stale and a new one replaced it.

For some reason, calls to watch-namespaces always fails when the resourceVersion is specified.

This behaviour causes the kubernetes/client-node informer class to go into an infinite loop trying to watch at a specific version. The fix is to omit the resourceVersion parameter (which was added to fix a different issue, it wasn’t added because we need to restrict to only the later versions, here 250f9e45).


jonnyowenpowell (approved):
LGTM 👍

PR #1571, opened 2025-03-26
chore: remove dead code

popas90:
- n/a Tests written and linted [ℹ︎](https://github.com/snyk/general/wiki/Tests)

### What this does

Remove dead code. Update contexts.


test/setup/platforms/index.ts
```diff
@@ -1,6 +1,4 @@
 import * as kind from './kind';
```
KatieArmstr, line 1: do we still run the kind tests?

popas90: not the integration ones. But the system tests still rely on kind when setting up:
https://github.com/snyk/kubernetes-monitor/blob/3a5ca525a5d18b5a5ec05b249f1154616e933f56/test/system/kind.spec.ts#L10

KatieArmstr (approved):

PR #1721, opened 2026-07-23
fix(deps): remove destabilizing ajv override and redundant overrides

parker-snyk:

### What this does

Cleans up the package.json overrides block, which has piled up entries over the years. Several are now redundant or out of date: the package each one pins resolves naturally to a version at or above what the override forces, so the override isn't doing anything anymore.

I dropped `ajv`, `ws`, `jsonpath-plus`, top-level `qs`, `fast-uri`, and `path-to-regexp`. Each resolves on its own to a safe, current version, so removing the override doesn't change what gets installed.

I kept the overrides that still do real work: `tough-cookie`, `form-data`, and `qs` under `@kubernetes/client-node` (pinning `request`'s old transitive deps up for vuln reasons), plus `tar`, `brace-expansion`, `adm-zip`, and the `@yarnpkg/parsers` js-yaml pin.

### Notes for the reviewer

No security regression. Snyk doesn't flag any of the removed packages, and each resolves to a version at or above the old override. Regenerating the lockfile also floats `js-yaml` to 4.3.0, which clears a high-severity js-yaml advisory.

I verified locally that `npm ci --omit=dev`, the `tsc` build, all 312 unit tests, and a full docker build all pass.

Tracking ticket: CN-1538 (in progress, assigned to me).

### Screenshots

N/A


ividalATSnyk (approved):
LGTM!

